PR #2, opened 2023-02-13
Sanity tests

spuiuk:
Make use of the new sanity test symlink in the sit-test-cases repo 
https://github.com/samba-in-kubernetes/sit-test-cases/pull/2

This is expected to be used by the CI environment in the sit-environment repo as we would like a quick turnaround when testing in the CI environment and we only need to validate the correctness of the sit-environment repo.

This was bought about because changes in the samba and clustered filesystem repo can cause failures in the smbtorture tests for no fault in the sit-environment repository.

To set up sanity testing only, we need to pass the 'test_sanity_only=1' in the EXTRA_VARS in the CI environment. 

depends on #pr1 
depends on samba-in-kubernetes/sit-test-cases#pr2


spuiuk:
closing this in favour of the approach in #3

PR #5, opened 2023-04-07
Failure when updating CentOS 8 vagrant vms.

spuiuk:
Changes in the rpm package specs required that the rpm package has to be updated first before we can updated the rest.

This doesn't seem to cause a problem in the CentOS environment which seems to use an updated centos8 image. However this does cause problems in devel environments when performing updates of the centos8 virtual machines.

### DIFF
--- a/vagrant/roles/local.defaults/tasks/centos8.yml
+++ b/vagrant/roles/local.defaults/tasks/centos8.yml
@@ -1,2 +1,2 @@
-vagrant_image: "centos/stream8"
+vagrant_image: "generic/centos8s"
 actions_file: "centos8.yml"


anoopcs9:
> Changes in the rpm package specs required that the rpm package has to be updated first before we can updated the rest.

ref: https://github.com/samba-in-kubernetes/samba-centosci/pull/5
 
> However this does cause problems in devel environments when performing updates of the centos8 virtual machines.

Actually updating RPM as such was my initial thought when I encountered the error. Later I realized that libvirt vagrant  boxes for CentOS Stream 8 are not always referenced to latest from CentOS. Also it is good to have the box version updated once in a while to avoid such problems. Therefore I suggest to update the box with more recent version than cluttering ansible roles/tasks with more workarounds.

spuiuk:
@anoopcs9, you are correct. The Centos/centos8 box hasn't been updated for over 2 years now.
https://app.vagrantup.com/centos/boxes/stream8

I am instead testing with boxes built by the roboxes.org team which seem to be updated far more frequently. 
https://app.vagrantup.com/generic/boxes/centos8s

This will improve the time taken as well since host updates should require fewer package updates. Will post the latest patch as soon I have confirmed that it works well on my machine. You may have to revert the centosci script changes once the new changes are merged.

anoopcs9:
> @anoopcs9, you are correct. The Centos/centos8 box hasn't been updated for over 2 years now. https://app.vagrantup.com/centos/boxes/stream8

I am aware of some efforts in place for automating this process but may not happen very soon.

> I am instead testing with boxes built by the roboxes.org team which seem to be updated far more frequently. https://app.vagrantup.com/generic/boxes/centos8s

I haven't tested it locally but I remember from the past that machines(generic/centos8) having issues while bringing up, assigning IP address and/or setting up SSH. But situation might have improved.

Having said that I am still in favour of manually updating the latest versions once in a while as they come officially from CentOS.

spuiuk:
> I haven't tested it locally but I remember from the past that machines(generic/centos8) having issues while bringing up, assigning IP address and/or setting up SSH. But situation might have improved.

It works fine. Also passed the github checks.

> Having said that I am still in favour of manually updating the latest versions once in a while as they come officially from CentOS.

This involves manually updating the URLs to point to the latest version every few months. While using the generic/centos8s image directly with Vagrant updates the images automatically.

anoopcs9:
> > I haven't tested it locally but I remember from the past that machines(generic/centos8) having issues while bringing up, assigning IP address and/or setting up SSH. But situation might have improved.
> 
> It works fine. Also passed the github checks.

Great.

> > Having said that I am still in favour of manually updating the latest versions once in a while as they come officially from CentOS.
> 
> This involves manually updating the URLs to point to the latest version every few months.

But it won't be for so long as I expect the automated process to land in coming months.

> While using the generic/centos8s image directly with Vagrant updates the images automatically.

Like CentOS, do we have responsive community and contact methods to ask questions on any future issues we may  encounter with those vagrant boxes?

spuiuk:
They have a website: https://roboxes.org/
along with a github project: https://github.com/lavabit/robox
with an active issues list.

Looking at the releases for Centos8s, They have been regular with about 22 updates in the last 11 months when it seems to have been started.
https://app.vagrantup.com/generic/boxes/centos8s

My concern with the boxes commands used is that it seems to be a more invasive change to add and remove later. Updating these images at the moment involves new commits to point to the latest versions and all of this is expected to change again in a next few months.
With this commit, when official centos boxes are released, the change required to point it to the official releases is trivial.

We can gather more opinions about this at the team call this evening.

anoopcs9 (approved):

anoopcs9:
Let's switch to `generic/centos8s` temporarily and revert once updated CentOS vagrant boxes are automatically referenced from https://app.vagrantup.com/centos/boxes/stream8.

PR #6, opened 2023-05-03
Modularize glusterfs-related tasks

xhernandez:
Remove most glusterfs-related tasks from main roles and group them into a new gluster specific role. Some tasks tightly bound to other roles have been kept in the same role but moved into one or more gluster specific task files.

The main purpose of this reorganization is to make it easier to add support for other backends, like ceph.


vagrant/ansible/roles/client.prep/tasks/centos8.yml
```diff
@@ -1,14 +1,3 @@
-- name: Install required packages
-  yum:
-    name:
-      - glusterfs-fuse
-      - cifs-utils
-      - samba-test
-      - python3-pip
-      - git
-      - make
```
phlogistonjohn, line 9: These seem pretty generic and I note that pip is used in the block immediately following this one. Maybe keep these few here?

xhernandez: You are right. Except for `glusterfs-fuse` it will be better to keep the other ones here. I'll update it.

anoopcs9:
```
PLAY [setup] *******************************************************************

TASK [Gathering Facts] *********************************************************
ok: [setup]

TASK [setup.prep : Run distro specific tasks for setup.prep] *******************
included: /root/tests/sit-environment/vagrant/roles/setup.prep/tasks/centos8.yml for setup

TASK [setup.prep : update packages] ********************************************
ok: [setup]

TASK [setup.prep : Enable EPEL repository] *************************************
ok: [setup]

TASK [setup.prep : Install Python3 pip] ****************************************
ok: [setup]

TASK [setup.prep : Install pip jinja2 library] *********************************
changed: [setup]

TASK [setup.prep : Install additional packages] ********************************
ok: [setup]

TASK [Process infrastructure specific tasks for setup] *************************
ERROR! 'filesystem' is undefined. 'filesystem' is undefined
make[1]: Leaving directory '/root/tests/sit-environment/vagrant'
make[1]: *** [Makefile:17: setup.prep.only] Error 1
make: *** [Makefile:3: test] Error 2
```

xhernandez:
@anoopcs9 sorry, I should have added a comment about this. For now the way to call it is to make use of `EXTRA_VARS` like this:

    EXTRA_VARS="filesystem=glusterfs" make

The PR is still a WIP and will be polished before final version. I just posted it to check whether the approach I've taken is the right one or not.

vagrant/ansible/roles/client.prep/tasks/centos7.yml
```diff
@@ -1,11 +1,11 @@
 - name: Install required packages
   yum:
     name:
-      - glusterfs-fuse
       - cifs-utils
       - samba-test
-      - python-pip
+      - python3-pip
```
anoopcs9, line 6: I think CentOS 7 still doesn't have Python 3+.

xhernandez: Yes. I messed it up when I moved them around. I'll fix it. Thanks.

xhernandez: I've just seen that the same file uses pip3 for the next task. Is that correct or should it be changed also ?

spuiuk:
> @anoopcs9 sorry, I should have added a comment about this. For now the way to call it is to make use of `EXTRA_VARS` like this:
> 
> ```
> EXTRA_VARS="filesystem=glusterfs" make
> ```

I think we should set a default value for the filesystem variable.

xhernandez:
> > @anoopcs9 sorry, I should have added a comment about this. For now the way to call it is to make use of `EXTRA_VARS` like this:
> > ```
> > EXTRA_VARS="filesystem=glusterfs" make
> > ```
> 
> I think we should set a default value for the filesystem variable.

In the latest version of the patches, it already uses a default value (for now 'glusterfs'), and the variable has been changed to `backend` instead of `filesystem`, which I think is better.

In the latest version you can use the `BACKEND` variable or the command `make glusterfs` to select the backend explicitly. Otherwise 'glusterfs' will be used.

xhernandez:
Is there any other concern or anything that should be changed for this PR ?

vagrant/ansible/roles/samba.setup/tasks/glusterfs/main.yml
```diff
@@ -0,0 +1,12 @@
+- name: Install samba and samba-vfs-glusterfs packages
```
anoopcs9, line 1: I think we can use the following centos8 specific glusterfs task as common to all by moving it under _samba.setup/tasks/glusterfs/main.yml_ (similar to what we have inside _ctdb.setup/tasks/glusterfs/main.yml_)

```
- name: Selinux - Allow Samba to access glusterfs services
  seboolean:
    name: "{{ item }}"
    state: yes
    persistent: yes
  with_items:
    - samba_load_libgfapi
    - samba_export_all_rw
```

I don't completely remember the need for `samba_export_all_rw` boolean but enabling it should not harm the test environment(centos7).

vagrant/ansible/roles/ctdb.setup/tasks/glusterfs/main.yml
```diff
@@ -0,0 +1,26 @@
+- name: SELinux - Allow CTDB to access recovery lockfile from FUSE mount
+  seboolean:
+    name: use_fusefs_home_dirs
+    state: yes
+    persistent: yes
+
+- name: Create the ctdb shared volume
```
anoopcs9, line 7: May be re-phrase to indicate that we also mount the volume created.

anoopcs9 (changes requested):
Very well done, nice re-organization.

See below for some review comments.

vagrant/ansible/roles/client.prep/tasks/centos7.yml
anoopcs9: Let's leave as it is for now 😐

vagrant/ansible/roles/sit.glusterfs/tasks/common.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Enable GlusterFS nightly rpms repository
+  command: yum-config-manager --add-repo http://artifacts.ci.centos.org/gluster/nightly/master.repo
+
+# required for dependencies
+- name: Enable GlusterFS repo
+  yum:
+    name: centos-release-gluster
+    state: latest
+
```
anoopcs9, line 19: new blank line at EOF 😛

vagrant/ansible/roles/ctdb.setup/tasks/main.yml
```diff
@@ -24,22 +20,21 @@
...
+- name: Process backend/distro-dependent tasks
+  include_tasks: "{{ include_file }}"
+  with_first_found:
+    - files:
+      - "{{ backend }}/{{ ansible_os_family }}.yml"
+      - "{{ backend }}/{{ distro.actions_file }}"
+      - "{{ backend }}/main.yml"
+      skip: True
+  loop_control:
+    loop_var: include_file
```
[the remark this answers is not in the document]
xhernandez: I've already done this (and some other related changes) in a new PR. Is it ok if we keep it this way here and check if the new PR is fine ?

vagrant/ansible/roles/samba.setup/tasks/glusterfs/main.yml
xhernandez: Done

vagrant/ansible/roles/sit.glusterfs/tasks/common.yml
xhernandez: Done

anoopcs9 (dismissed):
lgtm, thanks

vagrant/ansible/roles/client.prep/tasks/main.yml
```diff
@@ -2,6 +2,11 @@
   include_tasks:
     file: "{{ distro.actions_file }}"
 
+- name: Process infrastructure specific tasks for clients
+  include_role:
+    name: 'sit.{{ filesystem }}'
```
spuiuk: I checked the rest of the commit and couldn't find where the filesystem was define but from your earlier comments for this issue, I realised that we have to pass a new EXTRA_VARS for this to work properly. 
This was later renamed backend and the EXTRA_VARS is automatically set in the makefile. 

There are a few issues I have with this.
1) There will be a few commits where we cannot simply do a make test and it just works. This isn't much of an issue with this project and maybe we can ignore for this case. But I like to see continuity in patches.
2) Why not set backend in vagrant/roles/local.defaults/tasks instead of the Makefile? This would be consistent with how we set the distro variable.

spuiuk (changes requested):
Thanks for looking into this. Most of the commits look good. I just have a few minor comments and I am not too sure of the pip command. Let me know if I am wrong here.

vagrant/roles/sit.glusterfs/tasks/setup.yml
```diff
@@ -0,0 +1,40 @@
...
+      shell: ansible-galaxy collection install {{ item }}
+      loop:
+        - ansible.posix
+        - gluster.gluster
+        - ovirt.ovirt
+
+    - name: Install ansible collection dependencies
+      pip:
+        name: jmespath
+        executable: pip3.8
```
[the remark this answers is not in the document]
xhernandez: Done

vagrant/ansible/roles/client.prep/tasks/main.yml
xhernandez: > I checked the rest of the commit and couldn't find where the filesystem was define but from your earlier comments for this issue, I realised that we have to pass a new EXTRA_VARS for this to work properly. This was later renamed backend and the EXTRA_VARS is automatically set in the makefile.
> 
> There are a few issues I have with this.
> 
>     1. There will be a few commits where we cannot simply do a make test and it just works. This isn't much of an issue with this project and maybe we can ignore for this case. But I like to see continuity in patches.

I can squash all the commits into a single one if you prefer. I decided to make the changes once reviews had started so I thought that it would be easier for reviewers to apply the changes as new commits, but I'm ok to convert it to a single commit.

> 
>     2. Why not set backend in vagrant/roles/local.defaults/tasks instead of the Makefile? This would be consistent with how we set the distro variable.

As I see it, the idea is to make it very flexible by easily switching the OS or backend. Defining values inside a yaml of a role doesn't seem very flexible to me. In fact, in another PR I'm preparing I've moved the distro selection outside the role and it could be selected by environment variables or directly from `make` command like this:

    # make gf c7 (Build the test environment on CentOS 7 using GlusterFS)
or

    # make ceph c8 (Build the test environment on CentOS 8 using CephFS)

Once built, the configuration is stored and a simple `make test` should work without explicitly specifying the environment.

Is that ok ?

spuiuk: 1) I prefer to keep it as separate commits. So I guess we can leave it as it is for now.

2) By setting the field in vagrant/roles/local.defaults/tasks, you just set the default values to use if the value isn't defined in the EXTRA_VARS. 
example: To test a centos 7 build, we use
EXTRA_VARS="use_distro=centos7" make
This overrides the use_distro mentioned in the local.defaults.
see: https://github.com/samba-in-kubernetes/sit-environment/blob/main/docs/Fedora-Setup.md

xhernandez: From outside (being the first time one looks into this repo) it seems easier to find where default values are defined if they are near the command the user executes (i.e. inside the `Makefile` that `make` loads). Identifying that the defaults are set in a conditional task (not even a variables file) of a specific role two levels below the main directory is not obvious. That's why I didn't use the same approach for `backend`.

In any case, I'll modify it to use the same approach now.

vagrant/ansible/Makefile
```diff
@@ -1,8 +1,8 @@
 INVENTORY := ./vagrant_ansible_inventory
 
-BACKEND ?= glusterfs
-
-ANSIBLE_EXTRA_VARS := -e 'backend=${BACKEND}'
+ifdef BACKEND
```
spuiuk, line 3: I don't think you need to make any changes to the vagrant/Makefile and ansible/Makefile. We will be able to pass the backend variable by simply calling the playbook in the following manner

EXTRA_VARS="use_distro=centos7 backend=glusterfs" make

This will keep the Makefile clean.


I can add a document mentioning the variables which can be set. in a separate PR.

spuiuk (changes requested):

vagrant/ansible/Makefile
xhernandez: I'm aware of how to set variables in Makefiles. But now suppose I want to support 2 things, like the OS and the backend, with 3 different OS and 3 backends. Will you need to create 16 Makefile targets (including the targets where none or just one of the components is specified), each explicitly setting all values in `EXTRA_VARS` ?

What will happen if we add more OSes, or backends, or options to enable/disable or configure components in the OS, samba or the backend ? the number of targets will grow exponentially pretty fast. Maybe the Makefile structure will be simple, but it will be unmaintainable and it will be easy to make mistakes.

Probably it would be possible to avoid the exponential grow, but at the cost of adding `ifdef` and other Makefile constructs, which is what you wanted to avoid. On the other hand, using just independent variables is easy to avoid the exponential grow.

In any case, the latest set of patches make use of `EXTRA_VARS`, so please take a look and tell me if there's any other thing that needs to be modified.

PR #7, opened 2023-06-20
client.prep: Add pytest for clients

spuiuk:
We also symlink /usrt/local/bin/pytest to
/usr/bin/pytest-3

This will be used as a test runner for sit-test-cases

### DIFF
--- a/vagrant/ansible/roles/client.prep/tasks/centos8.yml
+++ b/vagrant/ansible/roles/client.prep/tasks/centos8.yml
@@ -6,6 +6,7 @@
       - python3-pip
       - git
       - make
+      - python3-pytest
     state: latest
 
 - name: Install Python 3 modules
@@ -15,3 +16,9 @@
       - PyYAML
       - iso8601
     extra_args: --ignore-installed
+
+- name: Link to pytest
+  file:
+    src: /usr/bin/pytest-3
+    dest: /usr/local/bin/pytest
+    state: link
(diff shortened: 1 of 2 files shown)


vagrant/ansible/roles/client.prep/tasks/centos8.yml
```diff
@@ -15,3 +16,9 @@
       - PyYAML
       - iso8601
     extra_args: --ignore-installed
+
+- name: Link to pytest
+  file:
+    src: /usr/bin/pytest-3
+    dest: /usr/local/bin/pytest
+    state: link
```
xhernandez, line 24: This is common to both CentOS 7 and 8. Maybe it could go inside main.yml.

xhernandez (commented):
Shouldn't these changes go to `client.test.prep` role ?

vagrant/ansible/roles/client.prep/tasks/centos8.yml
spuiuk: You are right but the packages the sit-test-cases are dependent on are already installed at this location. The client.test.prep is used to prepare the sit-test-cases git repo for executing the tests and it makes sense to keep that separate.

The reason for adding the symlink here was that while centos8 and centos7 may not provide pytest, my workstation, a F38 machine, the python3-pytest package provides a pytest executable in the form of a symlink. So when we start using Centos 9, we may end up going this route anyways.

xhernandez: ok

anoopcs9 (approved):

xhernandez (approved):

PR #8, opened 2023-06-21
Create dynamic configuration

xhernandez:
These PR introduces a dynamically created configuration file that can easily be tailored for different operating systems and backends.

This configuration will be used in a next PR to parametrize cluster settings (number of nodes, resources, ...)


anoopcs9:
First impression:

I was going through the changes commit by commit and https://github.com/samba-in-kubernetes/sit-environment/pull/8/commits/89d89f46682120eef6de2eb3a05f25a6a6bd758e seems like an overkill especially with all those **make** functions for conditionals and string substitution. Can we work towards a simplified approach?

xhernandez:
> First impression:
> 
> I was going through the changes commit by commit and https://github.com/samba-in-kubernetes/sit-environment/pull/8/commits/89d89f46682120eef6de2eb3a05f25a6a6bd758e seems like an overkill especially with all those **make** functions for conditionals and string substitution. Can we work towards a simplified approach?

I'm a bit surprised by this comment given the complexity of other Makefiles in other SINK repos. I think this is a very simplified way to achieve the changes I wanted. The only thing I can do is to remove it.

Is the use of SIT_OS and SIT_BE ok, or should I restrict it to EXTRA_VARS?

spuiuk:
Just want to point out that though the CI seems to have passed, it hasn't really. This is the testing result on the client.

--
+ EXTRA_VARS=test_sanity_only=1
+ make test
Makefile:10: warning: overriding recipe for target 'test'
Makefile:7: warning: ignoring old recipe for target 'test'
make: Circular test <- test dependency dropped.
--

https://jenkins-samba.apps.ocp.cloud.ci.centos.org/job/samba_gluster-integration-environment/152/consoleText

anoopcs9:
> > I was going through the changes commit by commit and [89d89f4](https://github.com/samba-in-kubernetes/sit-environment/commit/89d89f46682120eef6de2eb3a05f25a6a6bd758e) seems like an overkill especially with all those **make** functions for conditionals and string substitution. Can we work towards a simplified approach?
> 
> I'm a bit surprised by this comment given the complexity of other Makefiles in other SINK repos. I think this is a very simplified way to achieve the changes I wanted. The only thing I can do is to remove it.

nvm, you can fix the Makefile warnings and I'll go through it again.

vagrant/roles/setup.prep/tasks/main.yml
```diff
@@ -22,10 +22,18 @@
...
+    recurse: yes
+
 - name: copy ssh config in place
   copy:
     src: ansible/ssh-config-setup
     dest: /home/vagrant/.ssh/config
     mode: 0600
     owner: vagrant
     group: vagrant
+
```
anoopcs9, line 39: nit: new blank line at EOF.

anoopcs9 (approved):
lgtm.

You can remove the extra blank line(see below) if you happen to refresh the patch to address any other reviews. Otherwise just leave it.

xhernandez:
> lgtm.
> 
> You can remove the extra blank line(see below) if you happen to refresh the patch to address any other reviews. Otherwise just leave it.

If there are no other changes, I'll do it in next PR.

Would it make sense to fix all issues detected by yamllint ? many are for long lines (> 80 columns) and using yes/no instead of true/false. I'm not sure if we should be strict about that.

anoopcs9:
> Would it make sense to fix all issues detected by yamllint ? many are for long lines (> 80 columns) and using yes/no instead of true/false. I'm not sure if we should be strict about that.

How about creating an issue for further discussion? We could even think of adding the `yamllint` support as we do with other SINK repositories.

xhernandez:
> > Would it make sense to fix all issues detected by yamllint ? many are for long lines (> 80 columns) and using yes/no instead of true/false. I'm not sure if we should be strict about that.
> 
> How about creating an issue for further discussion? We could even think of adding the `yamllint` support as we do with other SINK repositories.

Done: #10

spuiuk (approved):
ACK

PR #12, opened 2023-07-24
Fix yamllint issues and provide an automatic check

xhernandez:


vagrant/roles/local.defaults/vars/main.yml
```diff
@@ -38,7 +38,7 @@ vagrant:
     private: 192.168.122.0/24
     public: 192.168.123.0/24
 
-  # Storage pool to use for the virtual disks (optional)
+#  # Storage pool to use for the virtual disks (optional)
```
anoopcs9, line 41: extra hash(`#`)?

.github/workflows/checks.yml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request:
+    branches: [main]
+
+jobs:
+  yamllint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run yamllint check
+        run: make yamllint
```
anoopcs9, line 14: Can we have the above change in a separate commit? Probably after another individual commit which adds the `yamllint` Makefile target?

anoopcs9 (changes requested):

vagrant/roles/local.defaults/vars/main.yml
xhernandez: No. This one is on purpose. The problem here is that comments in yaml are required to be put at the same indentation as the next line. This comment is related to the setting in the next line, so it's "re-commented" because the setting is also commented. The first hash is required to avoid a yamllint error. The second hash is left there so that if you uncomment both lines by removing the first hash, the comment will be correctly indented without requiring manual indentation.

I think this is the cleanest/easier way to avoid yamllint errors for these comments. The alternative would be to disable comment rule in `.yamllint` and allow any indentation for comments.

.github/workflows/checks.yml
xhernandez: Sure. I'll split the commits in the next update.

vagrant/roles/local.defaults/vars/main.yml
anoopcs9: Got it, thanks for the explanation.

spuiuk (approved):
ACK

xhernandez:
@phlogistonjohn you were more concerned about using a relaxed set of rules for yamllint. Are these rules ok to you ?

phlogistonjohn (approved):
I'm generally OK with choosing yamllint rules that make sense for each project, and those may be more relaxed than yamllint's defaults. Changes LGTM.

anoopcs9 (approved):

PR #16, opened 2023-08-02
Record backend into test-info.yml

spuiuk:
We can setup our tests based on the backend used.


xhernandez:
Shouldn't the backend be irrelevant from the client point of view ?

spuiuk:
In our experience running these tests against glusterfs, we came across several tests which failed. The number of failures in the test results added too much noise to the results and made it difficult to track changes in behaviour. 
To reduce the number of known faiures in testing, we created a separate list of tests (flapping.glusterfs in smbtorture) where we listed these tests as flapping results. This allowed us to continue working on these cases in the background while we continued running the nightly tests.

With the plans to add new backends to the sit-environment, we will come across more such tests which fail with specific backends. My plan is to use the backend information to craft a  list flapping test for each backend which contains these tests under investigation.

xhernandez (approved):

anoopcs9 (approved):

PR #19, opened 2023-08-04
client.test.prep: Change test directory to sit-test-cases

spuiuk:
Change the directory in which the sit-test-cases are cloned from samba-integration-tests to
sit-test-cases

### DIFF
--- a/vagrant/Makefile
+++ b/vagrant/Makefile
@@ -41,7 +41,7 @@ client.test:
 setup.site: setup.cluster setup.clients generate.report client.test
 
 client1.test:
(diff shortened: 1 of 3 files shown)


xhernandez (changes requested):
There's another reference to "samba-integration-tests" directory in `vagrant/Makefile`

spuiuk:
> There's another reference to "samba-integration-tests" directory in `vagrant/Makefile`

Good catch. The target in the Makefile is mostly used by developers to trigger tests from host system.

anoopcs9 (approved):

PR #23, opened 2023-09-06
Fix cephadm dependencies issues

xhernandez:
When cephadm is installed directly from the latest GitHub version, no dependency check is done. To avoid missing some required depencencies, cephadm is now installed from the same repo as the ceph-common package, which should be recent enough.

### DIFF
--- a/vagrant/ansible/roles/sit.cephfs/tasks/server/centos.yml
+++ b/vagrant/ansible/roles/sit.cephfs/tasks/server/centos.yml
@@ -1,8 +1,7 @@
 ---
-# Note: the storage sig repo and the ceph-common package are only required to
-#       be able to mount the ceph volume from the host. All other ceph-related
-#       binaries are installed from cephadm, which is downloaded directly from
-#       the main development branch.
+# Note: the repo and the ceph-common package are only required to be able to
+#       mount the ceph volume from the host. Cephadm will download all other
+#       required container images independently of the repository.
 
 - name: Identify latest ceph:{{ config.data.branch }} build
   uri:
@@ -11,7 +10,7 @@
             ref={{ config.data.branch }}&\
             flavor=default&\
             status=ready&\
-            distros=centos/8/x86_64&\
+            distros=centos/{{ config.os.version }}/x86_64&\
             sha1=latest"
     return_content: true
   register: ceph_shaman
@@ -34,3 +33,9 @@
       - lvm2
       - ceph-common
     state: present
+
+- name: Download cephadm
+  get_url:
+    url: "{{ ceph_shaman.json[0].url }}/noarch/cephadm"
+    dest: /root/cephadm
+    mode: 0755
--- a/vagrant/ansible/roles/sit.cephfs/tasks/server/main.yml
+++ b/vagrant/ansible/roles/sit.cephfs/tasks/server/main.yml
@@ -6,12 +6,6 @@
   loop_control:
     loop_var: include_file
 
-- name: Download cephadm
-  get_url:
-    url: "https://github.com/ceph/ceph/raw/{{ config.data.branch }}/src/cephadm/cephadm.py"
-    dest: /root/cephadm
-    mode: 0755
-
 - name: "Create SSH key-pairs"
   command: >
     ssh-keygen -q -f /root/.ssh/id_rsa -N '' -C "ceph-management@{{ inventory_hostname }}"


xhernandez:
@anoopcs9 can you check if it works with this change ?

vagrant/ansible/roles/sit.cephfs/tasks/server/centos.yml
```diff
@@ -32,5 +31,6 @@
     name:
       - podman
       - lvm2
+      - cephadm
```
anoopcs9, line 34: Looks like `cephadm`  is a _noarch_ package and we haven't configured it yet. Can we add it to the baseurl field?

anoopcs9 (changes requested):

phlogistonjohn:
I'm curious, did you run into any missing dependencies when installing from github?

Also, I'm glad you removed the code to pull cephadm from github. Somehow I totally overlooked that you were doing that. The instructions on installing cephadm without using package repos is here: https://docs.ceph.com/en/latest/cephadm/install/#cephadm-install-curl

Trying to run cephadm from the single python file would have broken on main as of yesterday.

anoopcs9:
@phlogistonjohn 

> TASK [sit.cephfs : Pull containers] ********************************************

> fatal: [storage0]: FAILED! => {"changed": true, "cmd": ["/root/cephadm", "pull"], "delta": "0:00:00.162620", "end": "2023-09-06 06:08:00.164244", "msg": "non-zero return code", "rc": 1, "start": "2023-09-06 06:08:00.001624", "stderr": "Traceback (most recent call last):\n  File \"/root/cephadm\", line 40, in <module>\n    from cephadmlib.constants import (\nModuleNotFoundError: No module named 'cephadmlib'", "stderr_lines": ["Traceback (most recent call last):", "  File \"/root/cephadm\", line 40, in <module>", "    from cephadmlib.constants import (", "ModuleNotFoundError: No module named 'cephadmlib'"], "stdout": "", "stdout_lines": []}

> fatal: [storage2]: FAILED! => {"changed": true, "cmd": ["/root/cephadm", "pull"], "delta": "0:00:00.155260", "end": "2023-09-06 06:08:00.172173", "msg": "non-zero return code", "rc": 1, "start": "2023-09-06 06:08:00.016913", "stderr": "Traceback (most recent call last):\n  File \"/root/cephadm\", line 40, in <module>\n    from cephadmlib.constants import (\nModuleNotFoundError: No module named 'cephadmlib'", "stderr_lines": ["Traceback (most recent call last):", "  File \"/root/cephadm\", line 40, in <module>", "    from cephadmlib.constants import (", "ModuleNotFoundError: No module named 'cephadmlib'"], "stdout": "", "stdout_lines": []}

> fatal: [storage1]: FAILED! => {"changed": true, "cmd": ["/root/cephadm", "pull"], "delta": "0:00:00.165403", "end": "2023-09-06 06:08:00.196911", "msg": "non-zero return code", "rc": 1, "start": "2023-09-06 06:08:00.031508", "stderr": "Traceback (most recent call last):\n  File \"/root/cephadm\", line 40, in <module>\n    from cephadmlib.constants import (\nModuleNotFoundError: No module named 'cephadmlib'", "stderr_lines": ["Traceback (most recent call last):", "  File \"/root/cephadm\", line 40, in <module>", "    from cephadmlib.constants import (", "ModuleNotFoundError: No module named 'cephadmlib'"], "stdout": "", "stdout_lines": []}

anoopcs9:
> The instructions on installing cephadm without using package repos is here: https://docs.ceph.com/en/latest/cephadm/install/#cephadm-install-curl

Cool. I have a better suggestion.

vagrant/ansible/roles/sit.cephfs/tasks/server/centos.yml
anoopcs9: Hm.. multiple baseurls is not an option as they are only searched in case of failures. Therefore we could directly download the `cephadm` binary instead of installing the package as follows:

```
- name: Download cephadm
  get_url:
    url: {{ ceph_shaman.json[0].url }}/noarch/cephadm
    dest: /root/cephadm
    mode: 0755
```

anoopcs9:
> > The instructions on installing cephadm without using package repos is here: https://docs.ceph.com/en/latest/cephadm/install/#cephadm-install-curl
> 
> Cool. I have a better suggestion.

```diff --git a/vagrant/ansible/roles/sit.cephfs/tasks/server/main.yml b/vagrant/ansible/roles/sit.cephfs/tasks/server/main.yml
index acbb921..69593aa 100644
--- a/vagrant/ansible/roles/sit.cephfs/tasks/server/main.yml
+++ b/vagrant/ansible/roles/sit.cephfs/tasks/server/main.yml
@@ -8,7 +8,7 @@
 
 - name: Download cephadm
   get_url:
-    url: "https://github.com/ceph/ceph/raw/{{ config.data.branch }}/src/cephadm/cephadm.py"
+    url: "{{ ceph_shaman.json[0].url }}/noarch/cephadm"
     dest: /root/cephadm
     mode: 0755
```

vagrant/ansible/roles/sit.cephfs/tasks/server/centos.yml
xhernandez: Won't this have the same potential problem of missing dependencies ?

anoopcs9: I don't think so. This is as per the [docs](https://docs.ceph.com/en/latest/cephadm/install/#cephadm-install-curl) @phlogistonjohn linked earlier. Only change here is that we download `cephadm` based on latest main branch and not from a release.

xhernandez: That's fine. I'll update the PR.

anoopcs9 (dismissed):
lgtm, thanks.

phlogistonjohn (commented):
I found one issue with this, and I'm glad I looked. It appears that the shaman url is using `distros=centos/8/x86_64` which would be centos 8. Now, if we're installing on Centos8 this is fine, but I want to bring up the slight discrepancy between this and the doc I linked, that is the doc points the user toward the centos 9 build.  This is because the build process updates the python shebang - the shebang on centos8 builds is fine for centos 8 but not for other systems. Similarly, the ubuntu builds shebang is only appropriate for ubutnu. However, the shebang generated by centos9 builds is fairly universal, so we documented that for the general case it's preferred to download cephadm from the centos/9 build dir.

xhernandez:
> I found one issue with this, and I'm glad I looked. It appears that the shaman url is using `distros=centos/8/x86_64` which would be centos 8. Now, if we're installing on Centos8 this is fine, but I want to bring up the slight discrepancy between this and the doc I linked, that is the doc points the user toward the centos 9 build. This is because the build process updates the python shebang - the shebang on centos8 builds is fine for centos 8 but not for other systems. Similarly, the ubuntu builds shebang is only appropriate for ubutnu. However, the shebang generated by centos9 builds is fairly universal, so we documented that for the general case it's preferred to download cephadm from the centos/9 build dir.

You are right. I did it initially as a test to see if it worked, but later I forgot to fix it. My idea is to split the file into `centos7.yml` and `cento8.yml` to use the right value for each one. Once we add support for CentOS 9, we'll just need to add a `centos9.yml` with `centos/9/x86_64`.

Is that fine ?

phlogistonjohn:
> You are right. I did it initially as a test to see if it worked, but later I forgot to fix it. My idea is to split the file into `centos7.yml` and `cento8.yml` to use the right value for each one. Once we add support for CentOS 9, we'll just need to add a `centos9.yml` with `centos/9/x86_64`.
> 
> Is that fine ?

*shrug*  I can't much say if it's right or wrong, it is going to depend on what you intend to do. I just wanted to point out that difference between the doc and what the code is doing :-)

xhernandez:
I've created PR #27 to simplify this. I'll update this PR using the new properties from the other PR.

anoopcs9 (approved):

xhernandez:
Just rebased the PR

PR #24, opened 2023-09-06
Add GPFS backend

xhernandez:
Depends on #87


diederichm (commented):
I would suggest to remove the protocols -i eth0. This is only required for "interface" mode, and in standard setups CES public IPs for SMB access will find the interface by  looking for a matching subnet on one of the interfaces

xhernandez:
> I would suggest to remove the protocols -i eth0. This is only required for "interface" mode, and in standard setups CES public IPs for SMB access will find the interface by looking for a matching subnet on one of the interfaces

Ok. I'll do that.

vagrant/ansible/cluster-spectrumscale.yml
```diff
@@ -0,0 +1,30 @@
...
+    config.nodes |
+    dict2items |
+    selectattr('value.groups', 'contains', 'protocol') |
+    map(attribute='value.ctdb.public') |
+    list
+  }}
+ctdb_network_public_interface_subnet_mask: "24"
+ctdb_network_public_interface_name: "eth2"
+
+samba_netbios_name: "GLUSTERTEST"
```
phlogistonjohn, line 22: "GLUSTERTEST"

vagrant/ansible/roles/sit.storagescale/tasks/common/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Process OS specific tasks
+  include_tasks: "{{ include_file }}"
+  with_first_found:
+    - files: "{{ config.os.includes }}"
+  loop_control:
+    loop_var: include_file
+
+# - name: Disable rpcbind
```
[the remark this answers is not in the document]
xhernandez: The Storage Scale documentation recommends to disable rpcbind, so I added it as a step. However CentOS8 doesn't have it installed by default, which causes the task to fail.

I wasn't sure if it was better to keep it for reference or remove it (or make it a bit more complex to detect if rpcbind is actually installed).

Probably I can just remove it. I'll do it in the next update.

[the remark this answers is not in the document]
Shwetha-Acharya: we can even add one more commented line explaining this situation which can be used in future, so that this recommendation by scale does not go out of our radar (if that sound good to you.)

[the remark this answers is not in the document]
spuiuk: or we can ask ansible to ignore any failure. ie. handle both cases where rpcbind is installed or when not installed.

playbooks/ansible/roles/sit.gpfs/tasks/new_share.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Bind the GPFS volume to mount path
+  mount:
+    path: "{{ config.paths.mount }}/{{ volume }}"
+    src: "/gpfs/sit_fs/{{ volume }}"
+    opts: bind
+    fstype: none
+    state: mounted
```
anoopcs9, line 8: IMO above bind mount task is better fit for _new_volume.yml_.

playbooks/ansible/roles/sit.gpfs/tasks/new_share.yml
```diff
@@ -0,0 +1,18 @@
...
+
+- name: Set permissions to root directory
+  file:
+    path: "{{ config.paths.mount }}/{{ volume }}"
+    mode: 0777
```
anoopcs9, line 13: Isn't this already part of _fix_share_permissions.yml_ above?

playbooks/roles/sit.gpfs/tasks/setup/redhat.yml
```diff
@@ -0,0 +1,12 @@
...
+  yum:
+    name:
+      - ansible-core
+      - ansible-collection-ansible-posix
+      - ansible-collection-community-general
+      - unzip
+    state: latest
+
+- name: Install required ansible collections
+  command: ansible-galaxy collection install -p /usr/share/ansible/collections community.crypto
```
anoopcs9, line 12: Doesn't it work with the default installation path?

playbooks/ansible/roles/sit.gpfs/tasks/server/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Disable firewall
+  service:
+    name: firewalld
+    enabled: false
+    state: stopped
```
anoopcs9, line 6: Is this a requirement from Storage Scale?

playbooks/ansible/roles/sit.gpfs/tasks/common/redhat.yml
```diff
@@ -0,0 +1,14 @@
...
+      - unzip
+      - python3
+      - kernel-devel
+      - kernel-headers
+      - gcc
+      - gcc-c++
+      - elfutils
+      - elfutils-devel
+      - make
+    state: latest
```
anoopcs9, line 14: Aren't these better suited as _server_ tasks than _common_?

anoopcs9 (changes requested):

playbooks/ansible/roles/ctdb.setup/tasks/gpfs/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: 'sit.{{ config.be.name }}'
+    tasks_from: new_volume
+  vars:
+    volume:
+      name: "ctdb"
+
+- name: Create lock directory
+  file:
+    path: "{{ config.paths.ctdb.var_lib }}/lock"
+    state: directory
```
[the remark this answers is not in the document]
xhernandez: Correct. I'll change it.

playbooks/ansible/roles/ctdb.setup/tasks/gpfs/main.yml
```diff
@@ -0,0 +1,21 @@
...
+
+- name: Bind CTDB directory
+  mount:
+    path: "{{ config.paths.ctdb.var_lib }}/lock"
+    src: /gpfs/sit_fs/ctdb
+    opts: bind
+    fstype: none
+    state: mounted
```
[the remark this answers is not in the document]
xhernandez: Yes, that seems better. However I've seen that there's no consistency on how the variables are used to indicate that the volume needs to be mounted. I'll move the mount to _new_volume.yml_ but I'll leave for a future patch normalizing how variables are used.

playbooks/ansible/roles/ctdb.setup/tasks/main.yml
```diff
@@ -32,14 +32,22 @@
...
-    service: ctdb
-    permanent: yes
-    state: enabled
-
-- name: Reload firewall
-  command: firewall-cmd --reload
+- name: Check firewall
+  command: firewall-cmd --state
+  failed_when: false
+  register: fwd_state
```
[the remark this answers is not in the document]
xhernandez: Yes. This has been necessary because firewall configuration for GPFS is not trivial for each machine type and I preferred to implement it in a future patch, so GPFS disables the firewall. However, trying to configure the firewall when it's disabled, causes a failure, so I only do that if the firewall is actually enabled.

playbooks/ansible/roles/sit.gpfs/tasks/new_share.yml
xhernandez: I agree. To not make this too hacky in the current approach, I'll also make a change where the volume for the shares is created from _new_share.yml_ instead of the backend installation. This also seems more logical to me.

playbooks/ansible/roles/ctdb.setup/tasks/gpfs/main.yml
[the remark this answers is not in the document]
xhernandez: On second thought, that is not so simple. _new_volume.yml_ is called once from one of the nodes because the filesystem is distributed and it only needs to be created once. However the bind mount needs to be done on all samba nodes. This is also problematic when creating a share. Ideally, the creation of a share should include the creation of the volume, but the target nodes of both operations are also different.

If you are ok, I would prefer to leave this as it is now and try to normalize it for all backends in a future change.

playbooks/ansible/roles/sit.gpfs/tasks/new_share.yml
xhernandez: Check my previous comment. Bind needs to be done on all server nodes, but _new_volume.yml_ is only called once.

playbooks/ansible/roles/sit.gpfs/tasks/new_share.yml
xhernandez: Yes. I'll remove it.

playbooks/ansible/roles/sit.gpfs/tasks/server/main.yml
xhernandez: No, but I would prefer to address firewall configuration in another patch. To configure it correctly, each node requires a different set of open ports, depending on what services are running and what role has each node inside the Storage Scale cluster. With the current approach, where some playbooks are only run on subsets of the servers, making the firewall configuration correctly could be a bit of a mess.

playbooks/ansible/roles/sit.gpfs/tasks/common/redhat.yml
xhernandez: Yes. I confused the client nodes as the nodes accessing the backend, which require these components, but here "client" means Samba client, so it's backend agnostic and only requires samba client components. I'll change that.

playbooks/roles/sit.gpfs/tasks/setup/redhat.yml
xhernandez: By default it installs collections into the user's private directory (_~/.ansible/collections/ansible_collections_)

anoopcs9: I can confirm that `community.crypto` ansible collection is available for install from epel which should be enabled in _setup_ machine. May be we could try installing **_ansible-collection-community-crypto_** instead?

xhernandez: The package is not available for CentOS 8.

playbooks/settings.yml
```diff
@@ -350,3 +354,57 @@ environments:
...
+        ctdb:
+          public: 20
+        groups: [cluster, installer, gui, admin]
+        accounts: [default]
+
+      client:
+        cpu_weight: 1
+        memory_weight: 1
+        os: centos8
+        memory: 1024
```
anoopcs9, line 393: `gpfs.memory` above is already set as 1024. Do we still redefine it here?

playbooks/ansible/cluster-gpfs.yml
```diff
@@ -0,0 +1,29 @@
...
+
+samba_netbios_name: "SIT-GPFS-TEST"
+
+samba_shares:
+  - name: share
+    samba:
+      options:
+        "acl_xattr:ignore system acls": "yes"
+
+nsd_device: "/dev/vdb"
```
anoopcs9, line 29: Isn't it better to include this as `gpfs.nsd_device` under `samba_shares`?

anoopcs9 (changes requested):

playbooks/roles/sit.gpfs/tasks/setup/redhat.yml
anoopcs9: Is it OK to add `become: no` to avoid specifying the installation path? See https://github.com/samba-in-kubernetes/sit-environment/blob/main/playbooks/roles/sit.glusterfs/tasks/setup/centos.yml#L16

playbooks/settings.yml
xhernandez: It's not necessary. I'll remove it.

playbooks/ansible/cluster-gpfs.yml
xhernandez: Yes. I'll do that.

xhernandez: On second thought, it can't be defined inside the shares because the GPFS volume is global, and each share only creates a subvolume inside the already existing global volume.

playbooks/roles/sit.gpfs/tasks/setup/redhat.yml
xhernandez: With latest changes, playbooks on setup0 node are executed as root now, so I'm not sure if this is still relevant. I'll try without the explicit path.

playbooks/ansible/cluster-gpfs.yml
anoopcs9: Ok, thanks for the clarification.

anoopcs9:
A successful run: https://jenkins-samba.apps.ocp.cloud.ci.centos.org/job/samba_gpfs-integration-environment/9/

anoopcs9 (approved):

PR #29, opened 2023-09-08
Remove CentOS 7 support

xhernandez:


anoopcs9 (changes requested):
I could see few more references in the following files:

* _devel/Makefile_
* _devel/README.md_
* _docs/Fedora-Setup.md_

Also, _devel/centos7_build_setup.yml_ can also be removed.

anoopcs9 (approved):

PR #30, opened 2023-09-11
docs: start adjusting the Fedora setup guide for newer Fedora version 38

obnoxxx:
this not only mentions edora version 38 instead of  33 but also ads one more required package to be installed to run the environment on Fedora


gd (approved):
Great catch, I ran into the same issue recently as well.

PR #32, opened 2023-09-12
Add VFS support to CephFS backend

xhernandez:


vagrant/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
```diff
@@ -1,6 +1,9 @@
 [{{ name }}]
-comment = For samba share of volume {{ path }}
-vfs objects = acl_xattr
+comment = For samba share of subvolume {{ name }}
+vfs objects = acl_xattr{% if config.be.variant == 'vfs' %} ceph
```
phlogistonjohn, line 3: Minor style nitpick. Having an if block start in the middle of this line but continue over multiple lines looks really awkward to me. Perhaps just an if-else-endif would read better?

phlogistonjohn (commented):
Generally, looks like you're heading in the right direction to me.

vagrant/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
xhernandez: I'll try to make it prettier.

obnoxxx:
@xhernandez  wrote: 
> I'm not sure what's wrong, but most of the tests fail with error NT_STATUS_NO_MEMORY. Even increasing VM's memory to 4 GiB the error is the same. I'll try to investigate this a bit more.

This might already be a first indication that the cephfs vfs module has the same problem as the gluster module: 

for each share connection (not TCP connection), the entire cephfs client lib is instantiated to establish a ceph connection from userspace.  since the  lib is rather heavyweight, this can lead to memory problems when a lot of share connections are established by a test or other workload.

xhernandez:
> > I'm not sure what's wrong, but most of the tests fail with error NT_STATUS_NO_MEMORY. Even increasing VM's memory to 4 GiB the error is the same. I'll try to investigate this a bit more.
> 
> This might already be a first indication that the cephfs vfs module has the same problem as the gluster module:
> 
> for each share connection (not TCP connection), the entire cephfs client lib is instantiated to establish a ceph connection from userspace. since the lib is rather heavyweight, this can lead to memory problems when a lot of share connections are established by a test or other workload.

I don't understand exactly why, but the problem was that CTDB was checking the presence of the path of the samba shares. Since the path is not physically there, CTDB was not going up and clients were unable to connect. I've disabled this check and "only" 5 tests have failed.

I'm not sure how this translates into memory errors, but now they have disappeared.

phlogistonjohn:
There are parts of the code base that chooses some pretty odd error codes. It was probably due to that.

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
```diff
@@ -4,7 +4,29 @@
...
+        policy: targeted
+        state: permissive
+
+    - name: Skip check of share path
+      lineinfile:
+        path: /etc/ctdb/script.options
+        create: true
+        search_string: "CTDB_SAMBA_SKIP_SHARE_CHECK"
+        line: "CTDB_SAMBA_SKIP_SHARE_CHECK=yes"
+        state: present
```
anoopcs9, line 26: Good catch.

For the default VFS GlusterFS mode we always had `/` and it was not a problem. But with CephFS subvolumes share path can no longer be just `/`. Thanks for investigating the issue.

man [ctdb-script.options(5)](https://ctdb.samba.org/manpages/ctdb-script.options.5.html) also lists an option to have such script-specific options to be defined in a separate file i.e, _50.samba.options_ in this case.

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
```diff
@@ -4,7 +4,29 @@
...
+  block:
+    - name: Install CephFS VFS module
+      yum:
+        name: samba-vfs-cephfs
+        state: present
+
+    - name: Set selinux in permissive mode
+      selinux:
+        policy: targeted
+        state: permissive
```
anoopcs9, line 18: Moving forward we will find ways(policies, allow rules, booleans etc) to make Samba-CephFS integration via VFS module work in default _Enforcing_ mode.

anoopcs9 (commented):
Good reorg and clean changes 👏🏻 . You found the right place for backend specific _smb.conf_ share definitions by moving those under respective roles.

Everything looks good to me, thanks.

anoopcs9:
> > > I'm not sure what's wrong, but most of the tests fail with error NT_STATUS_NO_MEMORY. Even increasing VM's memory to 4 GiB the error is the same. I'll try to investigate this a bit more.
> > 
> > 
> > This might already be a first indication that the cephfs vfs module has the same problem as the gluster module:
> > for each share connection (not TCP connection), the entire cephfs client lib is instantiated to establish a ceph connection from userspace. since the lib is rather heavyweight, this can lead to memory problems when a lot of share connections are established by a test or other workload.
> 
> I don't understand exactly why, but the problem was that CTDB was checking the presence of the path of the samba shares. Since the path is not physically there, CTDB was not going up and clients were unable to connect. I've disabled this check and "only" 5 tests have failed.

Look what I found: https://wiki.samba.org/index.php/Samba_CTDB_GlusterFS_Cluster_HowTo#The_file_script.options. People already knew but we haven't had any problems reported with GlusterFS as not many might have used subdir based shares while setting up Samba.

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
xhernandez: I've updated the PR to use the script specific options, but the path of that file can be considered stable ?

vagrant/roles/local.defaults/tasks/main.yml
```diff
@@ -5,7 +5,8 @@
     dest: ./ansible/config.yml
   vars:
     os: "{{ use_distro | default('centos8') }}"
-    be: "{{ backend | default('glusterfs') }}"
+    be: "{{ (backend | default('glusterfs') | split('.'))[0] }}"
+    variant: "{{ (backend | default('glusterfs') | split('.'))[1] | default('default') }}"
```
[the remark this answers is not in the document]
xhernandez: Done

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
xhernandez: What's the correct path to add the _.options_ file ?

  - /etc/ctdb/events/legacy/
  - /usr/share/ctdb/events/legacy/

anoopcs9: > What's the correct path to add the _.options_ file ?
> 
> - /etc/ctdb/events/legacy/

This should be it. As the man page says:
...
_options for SCRIPT are placed in a file alongside the script, with a ".script" suffix added_
..

Those legacy events scripts we enable using [ctdb.setup](https://github.com/samba-in-kubernetes/sit-environment/blob/main/vagrant/ansible/roles/ctdb.setup/tasks/main.yml) role appear as symlinks under /etc/ctdb/events/legacy/.

> - /usr/share/ctdb/events/legacy/

 All available event scripts are located here.

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
```diff
@@ -4,7 +4,29 @@
...
+
+    - name: Skip check of share path
+      lineinfile:
+        path: /usr/share/ctdb/events/legacy/50.samba.options
```
anoopcs9, line 22: Proper location would be _/etc/ctdb/events/legacy/50.samba.options_.

anoopcs9 (changes requested):

vagrant/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
```diff
@@ -0,0 +1,14 @@
+[{{ name }}]
+comment = For samba share of subvolume {{ name }}
+vfs objects = acl_xattr
+{%- if config.be.variant == 'vfs' %} ceph
+ceph: config_file = /etc/ceph/sit.ceph.conf
+ceph: user_id = sit
```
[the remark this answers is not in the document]
xhernandez: Fixed

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
xhernandez: Done. Thanks

vagrant/ansible/roles/samba.setup/tasks/cephfs/main.yml
xhernandez: Done

anoopcs9 (approved):

phlogistonjohn (approved):
lgtm

PR #33, opened 2023-09-13
report.versions: Clean up and add Ceph version to the matrix

anoopcs9:


xhernandez (dismissed):

xhernandez (dismissed):

vagrant/ansible/roles/report.versions/tasks/main.yml
```diff
@@ -10,7 +10,16 @@
...
     - kernel
-    - samba-client
     - samba
-    - ctdb
     - glusterfs
+
+- name: Extract Ceph version
+  block:
+    - name: Run 'cephadm version'
+      shell: /root/cephadm version | awk -F"[()]" '{print $2}'
```
xhernandez, line 19: I've checked in my test machine and cephadm and ceph are the same version, but given how we install it, I'm not sure if this is guaranteed.

anoopcs9:
Here's how the version is reported as per the current set of changes:

> TASK [report.versions : debug] *************************************************
skipping: [setup0]
ok: [storage0] => {
    "msg": "ceph main-980a0a4f84ac21fb2fda40e61a048b808ec2c974"
}
ok: [storage1] => {
    "msg": "ceph main-980a0a4f84ac21fb2fda40e61a048b808ec2c974"
}
ok: [storage2] => {
    "msg": "ceph main-980a0a4f84ac21fb2fda40e61a048b808ec2c974"
}
skipping: [client0]

vagrant/ansible/roles/report.versions/tasks/main.yml
anoopcs9: I assume they are directly related i.e, the `cephadm` version used to setup a cluster will always match the version of Ceph that gets installed on the nodes. In case of any mismatch I expect a failure in cluster setup.

@phlogistonjohn can you please confirm?

phlogistonjohn: That's true unless you manually specify the image (`--image` option) to use during bootstrap. Otherwise whatever is "hard coded" in the cephadm sources will be used and that's generally going to match the build (main=>main, reef=>reef, etc).

xhernandez: My concern is about the exact hash (commit) we are testing. Cephadm is installed from a repo, while the containers are installed from an image in quay. I'm not sure if it's guaranteed that the repo and the quay image are always built from the same exact commit.

In case of a problematic patch, having it or not may be significant, and we should know it for sure.

phlogistonjohn:
Generally looks OK to me modulo the debug commit. Remove it and I'll happily approve.

vagrant/ansible/roles/report.versions/tasks/main.yml
anoopcs9: How about running `cephadm shell -- ceph version` instead of `cephadm version`? This will be more helpful I guess.

xhernandez: I think that the best solution could be to show the version of cephadm **and** ceph.

anoopcs9: OK. Updated.

xhernandez:
It looks good to me.

anoopcs9:
This is how it will be reported after including the suggestions:

> TASK [report.versions : debug] *************************************************
skipping: [setup0]
ok: [storage0] => {
    "msg": [
        "ceph main-38f11f28b05bfd80aaf2644cc1660ef8b51dd272",
        "cephadm main-38f11f28b05bfd80aaf2644cc1660ef8b51dd272"
    ]
}
ok: [storage1] => {
    "msg": [
        "ceph main-38f11f28b05bfd80aaf2644cc1660ef8b51dd272",
        "cephadm main-38f11f28b05bfd80aaf2644cc1660ef8b51dd272"
    ]
}
ok: [storage2] => {
    "msg": [
        "ceph main-38f11f28b05bfd80aaf2644cc1660ef8b51dd272",
        "cephadm main-38f11f28b05bfd80aaf2644cc1660ef8b51dd272"
    ]
}
skipping: [client0]

anoopcs9:
> Generally looks OK to me modulo the debug commit. Remove it and I'll happily approve.

Removed and updated.

xhernandez (approved):

anoopcs9:
Ignoring DNS related failures for **centos-ci/cephfs** as they are unrelated to the change. There is a solution proposed in #38.

PR #39, opened 2023-09-20
cephfs: remove unnecessary MDS creation

xhernandez:


anoopcs9 (approved):

spuiuk:
The changes look fine but can you give me, a cephfs newbie, an explanation for why we don't need this step here.  Does creating a volume automatically create a MDS?

xhernandez:
> The changes look fine but can you give me, a cephfs newbie, an explanation for why we don't need this step here.  Does creating a volume automatically create a MDS?

Yes. When the volume is created an MDS for that volume is also created. The previous command just created an MDS for an hypothetical volume named cephfs, but it even didn't exist. So the final result was a single filesystem _sit_fs_, but two MDS: one for _sit_fs_ and one for _cephfs_.

spuiuk (approved):
ACK

phlogistonjohn (approved):
LGTM

PR #41, opened 2023-09-21
Add mechanism to collect config and logs from nodes

anoopcs9:


vagrant/logs.collect.yml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts: cluster
```
xhernandez, line 2: This should be "all". Otherwise we won't have logs from client machines.

vagrant/roles/local.defaults/vars/main.yml
```diff
@@ -160,3 +160,7 @@ environments:
         networks:
           public: 5
         groups: [clients]
+
+misc:
+  logs:
+    basedir: "/tmp"
```
xhernandez, line 166: Just a minor comment, but I would move this upper in the configuration file. Having it here will make it less visible because the environment definitions are long.

vagrant/roles/logs.collect/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Create log directory structure
+  file:
+    path: "{{ config.logdir }}/{{ inventory_hostname }}"
+    state: directory
+  delegate_to: localhost
+
+- name: Process backend specific tasks
+  include_role:
+    name: 'sit.{{ config.be.name }}'
+    tasks_from: cluster/main.yml
```
xhernandez, line 11: I think that something like `collect` instead of `cluster` would be more meaningful.

xhernandez (commented):
Given that the machines are installed from scratch, logs should be small. Why we don't just copy all /var/log and journal contents ?

It's possible to miss important data, like kernel messages that don't explicitly mention ceph or any other component, audit data related to selinux, ...

Another thing that we should also collect are the configuration files. This can help detect problems in the playbooks themselves when generating those files.

vagrant/roles/local.defaults/vars/main.yml
anoopcs9: Done

vagrant/logs.collect.yml
anoopcs9: Done

vagrant/roles/logs.collect/tasks/main.yml
anoopcs9: I've split the tasks into `logs` and `config`.

anoopcs9:
> Given that the machines are installed from scratch, logs should be small. Why we don't just copy all /var/log and journal contents ?

Done.

> Another thing that we should also collect are the configuration files. This can help detect problems in the playbooks themselves when generating those files.

I've taken a basic approach to collect configuration from cluster nodes.

Please feel free to suggest changes to overall approach and nomenclature.

vagrant/roles/logs.collect/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+- name: Copy /var/log directory
+  synchronize:
+    src: /var/log
+    dest: "{{ config.statedir }}/{{ inventory_hostname }}"
+    mode: pull
+
+- name: Process backend specific tasks
+  include_role:
+    name: 'sit.{{ config.be.name }}'
+    tasks_from: logs/main.yml
```
xhernandez, line 17: I would move this before the copy of /var/log so that the backend has a chance to stop things correctly (if necessary), do whatever final checks are necessary, and move things to /var/log instead of copying them itself to the host.

Another alternative is to pass a variable with the destination to use, but I think that just requiring backends to store any logs in /var/log is cleaner and more flexible.

vagrant/roles/logs.collect/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- name: Copy journal
+  block:
+    - name: Extract journal since last boot
+      shell: journalctl -xb -n all > journal
```
xhernandez, line 22: Why not just extract it to /var/log/journal and move the task before copying /var/log ?

vagrant/roles/logs.collect/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Copy journal file
+      fetch:
+        src: journal
+        dest: "{{ config.statedir }}/{{ inventory_hostname }}/log/"
+        flat: true
+
+- name: Copy dmesg
+  block:
+    - name: Extract dmesg contents
+      shell: dmesg -T > dmesg
```
xhernandez, line 33: As in the previous comment, we can save it to /var/log/dmesg.kernel (for example) and let the other task to copy it.

vagrant/Makefile
```diff
@@ -35,6 +35,9 @@ setup.clients:
 generate.report:
 	@$(VSSH) admin "${SSH_EXTRA_VARS} make -C /home/vagrant/ansible generate.report"
 
+nodes.statedump:
+	@ansible-playbook --inventory $(INVENTORY) ${ANSIBLE_EXTRA_VARS} nodes.statedump.yml
```
xhernandez, line 39: I guess this will be called from jenkins after finishing the job (either successfully or with errors), right ?

Shouldn't we also add an entry in the main Makefile ?

vagrant/nodes.statedump.yml
```diff
@@ -0,0 +1,16 @@
+---
+- hosts: all
+  become: yes
+  become_method: sudo
+  vars_files:
+    - ansible/config.yml
+  roles:
+    - logs.collect
+
+- hosts: cluster
```
xhernandez, line 10: Isn't there anything in the client side that could be interesting ?

Right now we only use sit-test-cases. I'm not sure if it's the case, but there could be some client side configuration for the tests.

vagrant/roles/config.collect/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Create config directory structure
```
xhernandez, line 2: This task should be `run_once`.

vagrant/roles/logs.collect/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Create log directory structure
```
xhernandez, line 2: This task should be `run_once`.

vagrant/roles/config.collect/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  copy:
+    src: ansible/config.yml
+    dest: "{{ config.statedir }}"
+  delegate_to: localhost
+  run_once: true
+
+- name: Process backend specific tasks
+  include_role:
+    name: 'sit.{{ config.be.name }}'
+    tasks_from: config/main.yml
```
xhernandez, line 18: Instead of delegating the task to copy data to the host, maybe it would be cleaner to create a directory in the VM, let the backend and other modules copy configuration data there, and then copy the entire directory to the host in a single step from this role.

vagrant/roles/logs.collect/tasks/main.yml
xhernandez: > This task should be `run_once`.

My bad. I didn't see that each VM has its own directory.

vagrant/roles/logs.collect/tasks/main.yml
anoopcs9: Done.

vagrant/roles/logs.collect/tasks/main.yml
anoopcs9: Done.

vagrant/roles/logs.collect/tasks/main.yml
anoopcs9: Done.

vagrant/Makefile
anoopcs9: > I guess this will be called from jenkins after finishing the job (either successfully or with errors), right ?

Yes, that's the plan.

> Shouldn't we also add an entry in the main Makefile ?

Yes, thanks for the reminder.
Added now.

vagrant/nodes.statedump.yml
anoopcs9: I have included a test configuration data file for now.

vagrant/roles/config.collect/tasks/main.yml
anoopcs9: Updated to have a `configdir` per VM to collect necessary configuration data.

vagrant/nodes.statedump.yml
```diff
@@ -0,0 +1,16 @@
...
+  roles:
+    - logs.collect
+
+- hosts: cluster:clients
+  become: yes
+  become_method: sudo
+  vars_files:
+    - ansible/config.yml
+  roles:
+    - config.collect
```
xhernandez, line 16: Maybe just adding the _config.collect_ role to the previous playbook is enough.

anoopcs9: Correct.

xhernandez (approved):

Shwetha-Acharya (approved):

PR #42, opened 2023-09-28
Modularize tests run in SITE

xhernandez:
This is a first step to be able to run other tests inside SITE besides sit-test-cases.

Depends on #48


vagrant/ansible/roles/tests/tasks/log.yml
```diff
@@ -0,0 +1,16 @@
...
+  delegate_to: localhost
+  become: true
+  block:
+    - name: Get current time
+      command: date "+%Y-%m-%d %H-%M-%S.%6N"
+      register: now
+
+    - name: Write log line
+      lineinfile:
+        dest: /var/log/site-tests.log
```
anoopcs9, line 12: How about having separate log files for different tests? May be 
```suggestion
        dest: /var/log/log.{{ role }}
```

With #41 we could also think about having it under `{{ config.statedir }}`:
```suggestion
        dest: {{ config.statedir }}/log.{{ role }}
```

xhernandez: I'm ok with using `config.statedir`. However I don't see the benefit of having a different log file for each test. This log is really small, and used just to know when each test started and ended, and if it failed or not (and when). The idea is to be able to track times from global logs and map them to specific tests if necessary. Note that a "test" here represents a black box that can do anything, like sit-test-cases, which does a lot of smaller tests internally, but those are not seen nor logged by the sit-environment.

Each test will be able to generate its own logs if necessary, and I plan to incorporate monitoring of cpu, memory and other resources which will be independent per test.

anoopcs9: Fine.

anoopcs9 (approved):

vagrant/ansible/roles/tests/tasks/log.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Log the line
```
phlogistonjohn, line 2: This is minor but instead of naming it after what it does, maybe name it after why it's doing this?

xhernandez: I don't understand what you are proposing here. This is the name of the task and it normally tells what the task does. I've checked many other tasks, even on official documentation and that's the "normal" pattern. The "why" requires a bigger scope than a task and probably it's even unknown to this file. It depends on how these tasks are included from other files.

phlogistonjohn: No worries. I thought it was vague, but I guess you are saying it's supposed to be vague.

xhernandez: I got confused about the "why". If you want more details about what it does, I could change it to something like "Add a log entry in site-tests.log". Please, feel free to provide any suggestion you like.

xhernandez:
Depends on #48

anoopcs9 (approved):

phlogistonjohn (approved):
lgtm

PR #44, opened 2023-10-05
Move all vagrant dependent things inside a role and provide a generic way to define other provisioners

xhernandez:
This patch removes most of the vagrant specific things from the main files and concentrates them on a new _provisioner.vagrant_ role that is responsible for provisioning and destroying the virtual machines.

The approach makes it possible to define other provisioners in the future, like a libvirt-based one, or using a cloud provider, or even using physical machines.

Depends on #48


xhernandez:
Depends on #48

anoopcs9 (approved):
Good reorg.

lgtm, thanks.

PR #47, opened 2023-10-10
Make OS configurable per host

xhernandez:


vagrant/roles/local.defaults/templates/config.yml.j2
```diff
@@ -27,6 +29,7 @@ config:
     {%- for idx in range(data.instances | default(1)) +%}
     {{ name }}{{ idx }}:
       groups: {{ data.groups }}
+      os: {{ data.os | default(environments[be].os | default(os)) }}
```
anoopcs9, line 32: Can we have the optional `os` parameter documented in _vagrant/roles/local.defaults/vars/main.yml_?

anoopcs9 (changes requested):
Perfect change, thanks.

See below for a suggestion.

vagrant/roles/local.defaults/templates/config.yml.j2
xhernandez: Sure

anoopcs9 (approved):

PR #49, opened 2023-10-13
sit.glusterfs/setup: Install latest available ansible version

anoopcs9:
Now that the required [fixes](https://github.com/gluster/gluster-ansible-infra/pull/136) are in for gluster-ansible we could use latest ansible packages.

### DIFF
--- a/vagrant/roles/sit.glusterfs/tasks/setup/centos.yml
+++ b/vagrant/roles/sit.glusterfs/tasks/setup/centos.yml
@@ -5,9 +5,7 @@
 - name: Install gluster-ansible
   yum:
     name:
-      # Restricting to Ansible v2.12.x
-      # See https://github.com/gluster/gluster-ansible-infra/issues/135
-      - ansible-core-2.12*
+      - ansible-core
       - gluster-ansible-cluster
       - gluster-ansible-features
       - gluster-ansible-infra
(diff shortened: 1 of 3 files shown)


vagrant/roles/sit.glusterfs/tasks/setup/centos.yml
```diff
@@ -19,15 +17,14 @@
...
         - community.general
         - gluster.gluster
         - ovirt.ovirt
 
     - name: Install ansible collection dependencies
-      pip:
-        name: jmespath
-        executable: pip3.8
+      yum:
+        name: python3.11-jmespath
```
xhernandez, line 29: Didn't this had issues between centos8 and centos9 ?

anoopcs9: I thought about it while making the change. Should we split it now or at the time when we could successfully switch to CentOS 9 for GlusterFS cluster?

xhernandez: I think it's better to split it now. The playbook should be correct, even if we are not using part of it yet.

xhernandez: One thing we could do to avoid too much duplication is to leave the common part in centos.yml, and then include the task centos.yml from both centos8.yml and centos9.yml. After that we can just install the jmespath in the appropriate way for each centos version.

anoopcs9: Done.

vagrant/roles/sit.glusterfs/tasks/setup/centos8.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Process common OS tasks
+  include_tasks: "{{ config.os[config.nodes[inventory_hostname].os].distro }}.yml"
```
xhernandez, line 3: This file already has knowledge about the OS and version. It doesn't need to use `os` and `distro`. It can directly use the centos.yml file.

```suggestion
  include_tasks: centos.yml
```

vagrant/roles/sit.glusterfs/tasks/setup/centos9.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Process common OS tasks
+  include_tasks: "{{ config.os[config.nodes[inventory_hostname].os].distro }}.yml"
```
xhernandez, line 3: Same here.

xhernandez (approved):

PR #50, opened 2023-10-13
Add monitoring support

xhernandez:
Depends on #42.


xhernandez:
@anoopcs9 I'm wondering if this makes sense. sit-environment creates a virtual environment with small virtual disks, all running on the same physical host. This is fine to test functionality, but I'm not sure what real value could have statistics extracted from there.

What do you think ?

PR #54, opened 2023-11-02
statedump: Few cleanup changes

anoopcs9:
* Instead of processed backend name use the raw `backend` environment variable value for better discovery of statedir.
  - Such a change was the missing piece, especially with `cephfs.vfs` where jenkins used to search for _/tmp/sit.cephfs.vfs_statedump_ but the actual directory is named _sit.cephfs_statedump_.
* In those situations where environment was not setup correctly we fail to run any tests. This would mean that test-info.yml file is not linked and subsequent task to copy it fails. Therefore make sure that we collect rest of the logs and configuration by ignoring the errors while copying the test-info file.

### DIFF
--- a/playbooks/roles/local.defaults/templates/config.yml.j2
+++ b/playbooks/roles/local.defaults/templates/config.yml.j2
@@ -70,5 +70,5 @@ config:
     {%- endfor +%}
   {%- endfor +%}
 
-  statedir: "{{ misc.host.statedir }}/sit.{{ be }}_statedump"
+  statedir: "{{ misc.host.statedir }}/sit.{{ backend }}_statedump"
   configdir: "{{ misc.host.configdir }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/local.defaults/templates/config.yml.j2
```diff
@@ -70,5 +70,5 @@ config:
     {%- endfor +%}
   {%- endfor +%}
 
-  statedir: "{{ misc.host.statedir }}/sit.{{ be }}_statedump"
+  statedir: "{{ misc.host.statedir }}/sit.{{ backend }}_statedump"
```
xhernandez, line 73: Why not just use a constant name ? is it necessary to include the backend name ?

If we develop tools that automatically scan the data, it will need to know the backend before hand to easily access it. With a constant name, the tool can directly find the information and deduce the backed, if necessary, from the contents of the directory.

anoopcs9: > Why not just use a constant name ? is it necessary to include the backend name ?

I think its better to have it named after the backend used to easily differentiate in a first pass by just looking at the top-level directory name.

> If we develop tools that automatically scan the data, it will need to know the backend before hand to easily access it. With a constant name, the tool can directly find the information and deduce the backed, if necessary, from the contents of the directory.

Let's keep the backend(and/or variant) name until we have something along the lines mentioned above.

xhernandez (approved):

PR #55, opened 2023-11-02
Increase VM specs

xhernandez:
Upgraded number of CPUs and memory for all backends:

```
              Storage node       Client node
XFS          4 CPUs / 4 GiB    4 CPUs / 4 GiB
GlusterFS    8 CPUs / 8 GiB    4 CPUs / 4 GiB
CephFS       8 CPUs / 8 GiB    4 CPUs / 4 GiB
```

### DIFF
--- a/playbooks/roles/local.defaults/vars/main.yml
+++ b/playbooks/roles/local.defaults/vars/main.yml
@@ -107,17 +107,21 @@ provisioners:
 #
 environments:
   glusterfs:
-    cpus: 2
-    memory: 1024
+    cpus: 8
+    memory: 8192
 
     nodes:
       setup:
+        cpus: 2
+        memory: 1024
         networks:
           private: 200
         groups: [admin]
         os: centos8
 
       client:
+        cpus: 4
+        memory: 4096
         networks:
           public: 5
         groups: [clients]
@@ -134,11 +138,13 @@ environments:
         os: centos8
 
   xfs:
-    cpus: 2
-    memory: 1024
+    cpus: 4
+    memory: 4096
 
     nodes:
       setup:
+        cpus: 2
+        memory: 1024
         networks:
           private: 200
         groups: [admin]
@@ -158,14 +164,16 @@ environments:
         groups: [clients]
 
   cephfs:
-    cpus: 2
-    memory: 1024
+    cpus: 8
+    memory: 8192
 
     data:
       branch: main
 
     nodes:
       setup:
+        cpus: 2
+        memory: 1024
         networks:
           private: 200
         groups: [admin]
@@ -181,6 +189,8 @@ environments:
         groups: [cluster]
 
       client:
+        cpus: 4
+        memory: 4096
         networks:
           public: 5
         groups: [clients]


spuiuk:
Xavi, I like these changes but we need to keep in mind that these tests are also run on test machines with fewer resources than that available on the CentOS-CI systems.  I am inclined on ACKing these to get on with sit-test-cases development but we will need to prioritise the plans to vary the resources for test machines with fewer resources available.

spuiuk (approved):
ACK with certain conditions listed in an earlier comment.

xhernandez:
> Xavi, I like these changes but we need to keep in mind that these tests are also run on test machines with fewer resources than that available on the CentOS-CI systems. I am inclined on ACKing these to get on with sit-test-cases development but we will need to prioritise the plans to vary the resources for test machines with fewer resources available.

I'm thinking about implementing this approach:

Instead of setting a fixed number for the assigned memory to each node, we can provide a percentage of the free memory that the environment will use in a global setting. Then define the minimum memory required for each node and a weight. The weight will be used to distribute the remaining available memory (after subtracting the minimum required for each node) proportionally to each node.

For example:

```yaml
misc:
  host:
    memory: 50 # Percentage of the free memory to use
    ...
environments:
  glusterfs:
    memory: 1024 # Default minimum memory per node
    ...
    nodes:
      setup:
        memory_weight: 0 # 0 means to use the minimum memory
        ...
      client:
        memory_weight: 1
        ...
      storage:
        memory_weight: 2
        instances: 2
        ...
```

This configuration basically means that `setup` will always use 1024 MiB, and `storage` nodes will use twice the memory of the `client` nodes.

I was thinking about using this formula:

$$ B = \sum_{i=1}^n m_i $$

$$ T = \sum_{i=1}^n w_i $$

$$ M_k = m_k + \frac{FP - B}{T} w_k $$

Where:

  - $` n `$ is the total number of nodes that will be created.
  - $` m_i `$ is the minimum required memory for node $` i `$.
  - $` w_i `$ is the weight of node $` i `$.
  - $` F `$ is the amount of free memory.
  - $` P `$ is the percentage of free memory used for the environment.
  - $` M_i `$ is the amount of memory assigned to node $` i `$

With this, these are some possible configurations:

```
          58 GiB Free    48 GiB Free    38 GiB Free    28 GiB Free    18 GiB Free    8 GiB Free
setup0:      1 GiB          1 GiB          1 GiB          1 GiB          1 GiB         1 GiB
client0:     6 GiB          5 GiB          4 GiB          3 GiB          2 GiB         1 GiB
storage0:   11 GiB          9 GiB          7 GiB          5 GiB          3 GiB         1 GiB
storage1:   11 GiB          9 GiB          7 GiB          5 GiB          3 GiB         1 GiB
```

I would also truncate non-integer values to always use multiples of 1 GiB.

spuiuk:
While the approach makes sense, it would remove some of the flexibility available to the user to set a certain amount of resources tailored to their environment. I like the bare minimum default we have already set.

How about a simpler approach - Use an extra vars variable to indicate that the environment is a CentOS-CI test environment and therefore we can override the default memory settings with a higher number?

xhernandez:
I don't think it makes sense. Adding a flag to indicate if it's running in centos-ci or not will solve the problem only for centos-ci, but what happens to other setups where more resource are available ?

Even more, how will you manage the memory used for different node types ? clients don't require the same amount of memory as storage servers or even the setup machine. There could even be environments that have other types of machines with different requirements. This would require adding extra vars for each machine type. It's even worse if we also need to do that for CPUs (which I think we'll need to do).

I don't think this approach is simple nor scalable.

The approach I've proposed enforces a bare minimum. Using more memory just depends on the free memory (not just the total memory of the system) and the percentage configured. Everything else is dynamic and is controlled by each environment definition, so it's mostly transparent to the user.

IMO the only thing that could probably be good to override with an extra var is the percentage of memory to use by the environment.

spuiuk:
Let's go ahead with your approach. It should solve most of the problems we see. We can can revisit this if we hit roadblocks later.

anoopcs9 (approved):
Let the CI run with higher resources for couple of days. We can gradually include @xhernandez's logic from https://github.com/samba-in-kubernetes/sit-environment/pull/55#issuecomment-1792401262 to accommodate outside user environments.

PR #57, opened 2023-11-07
samba.setup: Ensure healthy cluster before `smbd` restart

### LINKED ISSUE #53: Cephfs: CI runs fail.
WRT to Cephfs
I have now seen multiple CI runs for PRs which fail with the following.

```
TASK [samba.setup : Restart samba] *********************************************
Thursday 19 October 2023  17:12:39 +0000 (0:00:00.684)       0:06:33.374 ****** 
fatal: [storage1]: FAILED! => {"changed": false, "msg": "Unable to start service smb: Job for smb.service canceled.\n"}
fatal: [storage0]: FAILED! => {"changed": false, "msg": "Unable to start service smb: Job for smb.service canceled.\n"}
fatal: [storage2]: FAILED! => {"changed": false, "msg": "Unable to start service smb: Job for smb.service canceled.\n"}
```

Eg: 
https://jenkins-samba.apps.ocp.cloud.ci.centos.org/job/samba_cephfs-integration-test-cases/122/console
https://jenkins-samba.apps.ocp.cloud.ci.centos.org/job/samba_cephfs-integration-test-cases/121/console

anoopcs9:
fixes #53

### DIFF
--- a/playbooks/ansible/roles/samba.setup/tasks/main.yml
+++ b/playbooks/ansible/roles/samba.setup/tasks/main.yml
@@ -49,6 +49,14 @@
   vars:
     name: '{{ item.share_name }}-{{ config.be.name }}-{{ config.be.variant }}'
 
+- name: Ensure cluster is in HEALTHY state
+  run_once: yes
+  command: ctdb nodestatus all
+  register: nodestatus
+  until: nodestatus.rc == 0
+  retries: 10
+  delay: 2
+
 - name: Restart samba
   service:
     name: smb


xhernandez (commented):
Wouldn't it be cleaner to do this after restarting ctdb ?

anoopcs9:
> Wouldn't it be cleaner to do this after restarting ctdb ?

Having it here would help us save time(few seconds) as other tasks can continue after CTDB restart. But on the other side checking for cluster healthiness is more suited to _ctdb.setup_ role. I'll include the suggestion for now.

xhernandez (dismissed):
I think this way is better. We already had several issues with ceph nodes starting in the background that caused unexpected failures in other tasks that were not so clearly related.

Not allowing things from running in the background makes it less likely that other tasks may fail and each ansible role is more self-contained.

anoopcs9:
> Wouldn't it be cleaner to do this after restarting ctdb ?

For now reverting back to older approach due to #58.

xhernandez (approved):

spuiuk (approved):
ACK - looks good.

PR #61, opened 2023-11-12
cluster-glusterfs: Disable disperse volume operations

anoopcs9:
Test runs on disperse volumes doesn't proceed after a hung state and eventually gets timed out(6 hrs from CentOS CI maximum reservation time). Therefore limit the tests runs to just distribute-replicated volumes which seems to pass without any issues.

### DIFF
--- a/playbooks/ansible/cluster-glusterfs.yml
+++ b/playbooks/ansible/cluster-glusterfs.yml
@@ -60,22 +60,6 @@ replicate_cluster_options:
   performance.readdir-ahead: 'on'
   performance.parallel-readdir: 'on'
 
-disperse_cluster_volume: "vol-disperse"
-disperse_cluster_disperse_count: 3
-disperse_cluster_bricks: '/bricks/brick-disp-0/vol,/bricks/brick-disp-1/vol,/bricks/brick-disp-2/vol'
-disperse_cluster_options:
-  features.cache-invalidation: 'on'
-  features.cache-invalidation-timeout: '600'
-  performance.cache-samba-metadata: 'on'
-  performance.stat-prefetch: 'on'
-  performance.cache-invalidation: 'on'
-  performance.md-cache-timeout: '600'
-  network.inode-lru-limit: '200000'
-  performance.nl-cache: 'on'
-  performance.nl-cache-timeout: '600'
-  performance.readdir-ahead: 'on'
-  performance.parallel-readdir: 'on'
-
 ctdb_network_private_interfaces: >-
   {{
     config.nodes |
@@ -104,4 +88,3 @@ samba_users:
 
 samba_shares:
   - { cluster_volume: "vol-replicate", share_name: "replicate" }
-  - { cluster_volume: "vol-disperse", share_name: "disperse" }
--- a/playbooks/ansible/roles/sit.glusterfs/tasks/new_volume.yml
+++ b/playbooks/ansible/roles/sit.glusterfs/tasks/new_volume.yml
@@ -7,18 +7,6 @@
     cluster: "{{ config.groups['cluster'] }}"
     replicas: "{{ volume.replica_count }}"
   run_once: yes
-  when: volume.type == 'replicate'
-
-- name: Create Dispersed Volumes
-  gluster_volume:
-    state: present
-    name: "{{ volume.name }}"
-    bricks: "{{ volume.backends }}"
-    cluster: "{{ config.groups['cluster'] }}"
-    disperses: "{{ volume.disperse_count }}"
-    force: yes
-  run_once: yes
-  when: volume.type == 'disperse'
 
 - name: Set volume options
   gluster_volume:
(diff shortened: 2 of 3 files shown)


playbooks/ansible/roles/sit.glusterfs/tasks/new_volume.yml
```diff
@@ -9,16 +9,16 @@
   run_once: yes
   when: volume.type == 'replicate'
```
xhernandez, line 10: If only one type of volume is created, this should be removed and we don't need to pass the `volume.type` argument.

playbooks/ansible/cluster-glusterfs.yml
```diff
@@ -60,21 +60,21 @@ replicate_cluster_options:
...
+#   features.cache-invalidation-timeout: '600'
+#   performance.cache-samba-metadata: 'on'
+#   performance.stat-prefetch: 'on'
+#   performance.cache-invalidation: 'on'
+#   performance.md-cache-timeout: '600'
+#   network.inode-lru-limit: '200000'
+#   performance.nl-cache: 'on'
+#   performance.nl-cache-timeout: '600'
+#   performance.readdir-ahead: 'on'
+#   performance.parallel-readdir: 'on'
```
xhernandez, line 77: Shouldn't we just delete this ?

If we want to re-add it in the future, it can be recovered from git history.

playbooks/ansible/roles/sit.glusterfs/tasks/new_volume.yml
anoopcs9: Sure.

playbooks/ansible/cluster-glusterfs.yml
anoopcs9: Fine.

xhernandez (approved):

spuiuk:
I don't see the point of continuing with replicated volume testing only if we don't intend on fixing problems we hit. Maybe we should probably switch off glusterfs testing in centos-ci instead. That way the code to do a more comprehensive test of glusterfs code is still available in easy reach.

spuiuk:
From the meeting today -
We investigate the cause of this problem to rule out issues in the test environment.
We don't merge this PR.
We run CentOS CI tests for glusterfs only for nightly runs and not for every PR.
We investigate if we can manually set a label(or other mechanism) to tell the CI system that the changes should also be tested against GlusterFS.

anoopcs9:
> From the meeting today - We investigate the cause of this problem to rule out issues in the test environment. We don't merge this PR.

Did you mean to investigate the hung state on running test suite for disperse volumes?

> We run CentOS CI tests for glusterfs only for nightly runs and not for every PR. We investigate if we can manually set a label(or other mechanism) to tell the CI system that the changes should also be tested against GlusterFS.

Above suggestion is already available for review as https://github.com/samba-in-kubernetes/samba-centosci/pull/42.

spuiuk:
> > From the meeting today - We investigate the cause of this problem to rule out issues in the test environment. We don't merge this PR.
> 
> Did you mean to investigate the hung state on running test suite for disperse volumes?

Yes. Xavi offered to investigate when he has some time. This can take some time now that we have disabled glusterfs testing with each PR.(See below)

> 
> > We run CentOS CI tests for glusterfs only for nightly runs and not for every PR. We investigate if we can manually set a label(or other mechanism) to tell the CI system that the changes should also be tested against GlusterFS.
> 
> Above suggestion is already available for review as [samba-in-kubernetes/samba-centosci#42](https://github.com/samba-in-kubernetes/samba-centosci/pull/42). Awaiting reviews..
Thanks - Acked and Mer

anoopcs9:
> > > From the meeting today - We investigate the cause of this problem to rule out issues in the test environment. We don't merge this PR.
> > 
> > Did you mean to investigate the hung state on running test suite for disperse volumes?
> 
> Yes. Xavi offered to investigate when he has some time. This can take some time now that we have disabled glusterfs testing with each PR.(See below)

Why don't we then take in the change so that nightly runs can complete on just distribute-replicate volumes without getting timed out?

xhernandez:
@anoopcs9 if we merge this PR, then there won't be any way to run disperse tests on centos-ci, right ?

To investigate the cause, I will need to be able to run some tests.

Is it possible to temporarily disable glusterfs tests on nightly builds ?

anoopcs9:
> > > From the meeting today - We investigate the cause of this problem to rule out issues in the test environment. We don't merge this PR.
> > 
> > 
> > Did you mean to investigate the hung state on running test suite for disperse volumes?
> 
> Yes. Xavi offered to investigate when he has some time. This can take some time now that we have disabled glusterfs testing with each PR.(See below)

https://github.com/samba-in-kubernetes/sit-test-cases/issues/45 will further track the progress on hung issue with disperse volumes. Therefore abandoning the PR as investigation is underway to root cause the hung state.

PR #62, opened 2023-11-13
devel: Allow modifying the setup to allow creation of single node samba accessing clustered filesystem

spuiuk:
When working on Samba to test against a clustered filesystem, I found it easier to use a single node samba server accessing a clustered filesystem in the backend. This cuts off ctdb and the clustered samba setup from the equation and allows us to focus on one samba server alone. We already have a test cluster setup with the samba-environment and can reuse this for development work.

I've added a couple of ansible playbooks in the devel tree to automate the process of disabling the samba cluster and setting up the development environment on storage0 of the cluster.

Instructions for ceph-vfs install which installs a ceph cluster on CentOS 9 servers and setting up a development environment on storage0 of the cluster:

1) Setup the test cluster
```
EXTRA_VARS="backend=cephfs.vfs" make -C playbooks/ setup.cluster setup.clients
```

2) (Optional step) fix the ssh keys on the test servers so that we can directly ssh into the servers. This also helps with rsync-ing files into the machine.
a) Copy a file "authorized_keys" containing the public keys you would like to copy to /root/.ssh/authorized_keys on the test machines.
b) run ansible playbook fix_ssh
```
make -C devel/ fix_ssh
```

3) Setup devel packages required to build samba on storage0 machine
```
make -C devel/  centos9_build_setup
```

4) Disable CTDB and stop the clustered samba service
```
make -C devel/ centos9_build_setup
```


xhernandez (approved):
I think the change is ok for quick developing/debugging purposes, but I think it would be better to integrate all these things into the main playbooks in the future. It would make much easier to correctly install debugging tools, sources and libraries specific for each backend, or create the same set of users for the local samba.

spuiuk:
> Looks OK to me. Longer term do you plan on making it possible to deploy a single service instance rather than undoing the ctdb setup?

At the moment there is no plan. It is something we could consider for the future but at the moment outside of a development environment we don't really have any need for it. Maybe if an when a need arises, we can consider it. Until then, this is an easier approach to set up a development environment.

PR #70, opened 2023-11-23
Make Samba and CTDB paths globally configurable

xhernandez:
The purpose of this patch is to centralize the location of different paths related to Samba and CTDB. This avoids differences between different backends, but also makes it possible to use the standard paths for them, which greatly reduces the friction with SELinux.


Shwetha-Acharya (approved):
LGTM

playbooks/ansible/roles/samba.setup/tasks/main.yml
```diff
@@ -1,7 +1,25 @@
...
+  yum:
+    name: samba
+    state: present
+
+- name: Create Samba share directories
+  file:
+    path: "{{ config.paths.mount }}/{{ item.cluster_volume }}"
+    state: directory
+    mode: 0777
+  loop: "{{ samba_shares }}"
```
anoopcs9, line 12: This means that we create a directory under _/mnt_ even if its unused for `glusterfs` and `cephfs.vfs`, right?

playbooks/ansible/roles/ctdb.setup/tasks/main.yml
```diff
@@ -14,14 +14,24 @@
...
 
+- name: Create lock directory
+  file:
+    path: "{{ config.paths.ctdb.var_lib }}/lock"
+    state: directory
+
+- name: Process backend specific tasks
+  include_tasks: "{{ config.be.name }}/main.yml"
+  vars:
+    ctdb_lock_path: "{{ config.paths.ctdb.var_lib }}/lock"
```
anoopcs9, line 30: I hope its OK to have mounts under _/var/lib/ctdb_.

playbooks/settings.yml
```diff
@@ -18,6 +18,17 @@ misc:
...
+paths:
+  samba:
+    etc: /etc/samba
+
+  ctdb:
+    var_lib: /var/lib/ctdb
+    etc: /etc/ctdb
+    bin: /bin
+
+  mount: /mnt
```
anoopcs9, line 30: Apart from `mount` all other path definitions seems a bit overkill to me. I would let each usage address the path directly.

playbooks/ansible/roles/samba.setup/tasks/main.yml
xhernandez: Yes. I'm even thinking to always mount them even if we later use a VFS module. It's always useful to have access to the data and it shouldn't cause any issue (if it does, it's a bug).

playbooks/ansible/roles/ctdb.setup/tasks/main.yml
xhernandez: I haven't seen any complaint from selinux and all tests worked fine. An alternative is to use a symlink, but this seemed to me more prone to trigger selinux issues.

playbooks/settings.yml
xhernandez: The problem is that we use it in many places. If for some reason we decide to move them elsewhere, changing the hardcoded paths will be harder. On the other side, this approach makes it much easier for the user to customize paths for their own environment (for example, installation path in /usr/local instead of the normal paths if samba and/or ctdb are installed from sources for developing/debugging reasons). It could even be used to automatically collect configuration settings from different components by automatically copying the _etc_ path from each component.

Having them on variables makes it possible to create a patch to easily override them from the `EXTRA_VARS` variable or other methods by the user.

playbooks/ansible/roles/samba.setup/tasks/main.yml
anoopcs9: Ok.

playbooks/settings.yml
anoopcs9: Thinking about source installations we could use the `prefix` approach so that standardized paths can be optionally prefixed with a user defined location.

Having said that we can start with this basic approach and see if in future it calls for the other approach.

PR #74, opened 2023-12-01
local.defaults: Change default backend to XFS

anoopcs9:
It has been a while since XFS was added to the list of supported backends and is known to successfully complete full test runs. On the other side GlusterFS is going through tough times in getting necessary fixes for long pending failures. Let's flip the switch and change the default backend from GlusterFS to XFS.


xhernandez (approved):

obnoxxx (approved):
LGTM

PR #76, opened 2024-01-16
Make definition of users and groups configurable inside settings.yml

xhernandez:
Instead of defining users inside the private configuration of each backend, the definition has been moved to the global _setting.yml_ file. Additionally, now it's also possible to define groups and group membership.

To support cases where we want a lot of users and/or groups, it's also possible to define them in a very compact way.

### DIFF
--- a/playbooks/ansible/roles/client.prep/templates/test-info.yml.j2
+++ b/playbooks/ansible/roles/client.prep/templates/test-info.yml.j2
@@ -8,6 +8,16 @@ exported_sharenames:
   {{ exported_sharenames | to_nice_yaml | indent(2) }}
 
 test_users:
-  {{ test_users | to_nice_yaml | indent(2) }}
+  {%- for name in config.accounts +%}
+    {%- for user_name in config.accounts[name].users +%}
+      {%- set user = config.accounts[name].users[user_name] +%}
+      {%- if user.samba +%}
+  - username: {{ user_name }}
+    uid: {{ user.uid }}
+    password: {{ user.password }}
+    nodes: {{ config.nodes | dict2items | selectattr('value.accounts', 'contains', name) | map(attribute='key') | list }}
+      {%- endif +%}
+    {%- endfor +%}
+  {%- endfor +%}
 
 test_backend: {{ test_backend }}
--- a/playbooks/roles/local.defaults/templates/config.yml.j2
+++ b/playbooks/roles/local.defaults/templates/config.yml.j2
@@ -20,6 +20,54 @@ config:
 
   paths: {{ paths }}
 
+  accounts:
+  {%- for name in accounts | default({}) +%}
+    {{ name }}:
+      groups:
+    {%- for group in accounts[name].groups | default({}) +%}
+      {%- set acct = accounts[name].groups[group] +%}
+      {%- for idx in range(acct.instances | default(1)) +%}
+        {%- set group_name = group +%}
+        {%- if acct.instances is defined +%}
+          {%- set group_name = group ~ (idx + 1) +%}
+        {%- endif +%}
+        {{ group_name }}: {{ acct.gid + idx }}
+      {%- endfor +%}
+    {%- endfor +%}
+      users:
+    {%- for user in accounts[name].users | default({}) +%}
+      {%- set acct = accounts[name].users[user] +%}
+      {%- for idx in range(acct.instances | default(1)) +%}
+        {%- set user_name = user +%}
+        {%- if acct.instances is defined +%}
+          {%- set user_name = user ~ (idx + 1) +%}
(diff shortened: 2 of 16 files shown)


playbooks/roles/local.defaults/templates/config.yml.j2
```diff
@@ -20,6 +20,54 @@ config:
...
+    {{ name }}:
+      groups:
+    {%- for group in accounts[name].groups | default({}) +%}
+      {%- set acct = accounts[name].groups[group] +%}
+      {%- for idx in range(acct.instances | default(1)) +%}
+        {%- set group_name = group +%}
+        {%- if acct.instances is defined +%}
+          {%- set group_name = group ~ (idx + 1) +%}
+        {%- endif +%}
+        {{ group_name }}: {{ acct.gid + idx }}
```
anoopcs9, line 34: Shouldn't we record `gid` as separate entity to be processed as `<group_name>.gid` further during tasks from _create_groups.yml_?

playbooks/ansible/roles/client.prep/templates/test-info.yml.j2
```diff
@@ -8,6 +8,16 @@ exported_sharenames:
...
 test_users:
-  {{ test_users | to_nice_yaml | indent(2) }}
+  {%- for name in config.accounts +%}
+    {%- for user_name in config.accounts[name].users +%}
+      {%- set user = config.accounts[name].users[user_name] +%}
+      {%- if user.samba +%}
+  - username: {{ user_name }}
+    uid: {{ user.uid }}
+    password: {{ user.password }}
+    servers: {{ config.nodes | dict2items | selectattr('value.accounts', 'contains', name) | map(attribute='key') | list }}
```
anoopcs9, line 18: Would `nodes` be more appropriate than `servers`?

anoopcs9 (changes requested):

playbooks/roles/local.defaults/templates/config.yml.j2
xhernandez: The groups don't require any additional information. In the settings file they are defined as an object with a `gid` field because we can also add the `instances` field, but in the config file we just need the number, so adding an object with a single field seems unnecessary. Instead of an object, I just use an integer.

In fact, the mistake is in _create_groups.yml_ where I use `gid` field when I shouldn't. I will fix that.

playbooks/ansible/roles/client.prep/templates/test-info.yml.j2
xhernandez: Done

playbooks/roles/local.defaults/templates/config.yml.j2
anoopcs9: Ok.

anoopcs9 (dismissed):

playbooks/settings.yml
```diff
@@ -29,6 +29,85 @@ paths:
 
   mount: /mnt
 
+# Creation of accounts:
+#
+# Here are defined the list of groups and users to create in the test nodes.
```
Shwetha-Acharya, line 34: Nit: defined the list -> defined list

playbooks/settings.yml
```diff
@@ -29,6 +29,85 @@ paths:
...
+#            incremented by one for each new group.
+#
+#       name: The name of the group will be suffixed by an incrementing index,
+#             starting from 1.
+#
+# Note: default group names (those with the same name as the user), don't need
+#       to be defined. They are created automatically. This object is intended
+#       to create supplementary groups.
+#
+# Users are created inside the 'users' subobject. Each user has the following
```
Shwetha-Acharya, line 60: same as above regarding subobject

playbooks/settings.yml
```diff
@@ -29,6 +29,85 @@ paths:
...
   mount: /mnt
 
+# Creation of accounts:
+#
+# Here are defined the list of groups and users to create in the test nodes.
+# Since it must be possible to define different accounts in different types of
+# machines, all definitions are created inside named objects, which will later
+# be referenced from the 'environment' object.
+#
+# Groups are created inside the 'groups' subobject. Each group has the following
```
Shwetha-Acharya, line 39: Nit: suboject -> sub object or sub-object

playbooks/settings.yml
xhernandez: Done

playbooks/settings.yml
xhernandez: Done

anoopcs9 (approved):

Shwetha-Acharya (approved):

PR #77, opened 2024-01-16
Make backend filesystems available from client machines

xhernandez:
This PR makes it possible to directly access the backend filesystem from the client machines bypassing Samba layer.

This is exclusively created for testing purposes since bypassing Samba will most probably cause problems (multiprotocol access is not supported yet), but it will simplify the implementation of some tests.

The backend filesystems are mounted inside _/mnt/backends/_, in a subdirectory with the same name as the samba share.

This is currently implemented using SSHFS, which doesn't support extended attributes, so there may be limitations on what can be done directly on the backend. If needed, this can be improved in the future by using an NFSv4 mount instead of SSHFS.

### DIFF
--- a/playbooks/ansible/cluster-glusterfs.yml
+++ b/playbooks/ansible/cluster-glusterfs.yml
@@ -44,38 +44,6 @@ gluster_infra_fw_zone: public
 gluster_infra_fw_services:
   - glusterfs
 
-replicate_cluster_volume: "vol-replicate"
-replicate_cluster_replica_count: "{{ config.groups['cluster']|length }}"
-replicate_cluster_bricks: '/bricks/brick-repl-0/vol,/bricks/brick-repl-1/vol,/bricks/brick-repl-2/vol'
-replicate_cluster_options:
-  features.cache-invalidation: 'on'
-  features.cache-invalidation-timeout: '600'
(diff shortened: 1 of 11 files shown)


playbooks/ansible/cluster-glusterfs.yml
```diff
@@ -100,5 +68,38 @@ ctdb_cluster_replica_count: "{{ config.groups['cluster']|length }}"
 samba_netbios_name: "SIT-GLUSTERFS-TEST"
 
 samba_shares:
-  - { cluster_volume: "vol-replicate", share_name: "replicate" }
-  - { cluster_volume: "vol-disperse", share_name: "disperse" }
+  - cluster_volume: "vol-replicate"
```
Shwetha-Acharya, line 71: just curious what is the advantage of this change?

spuiuk (dismissed):
ACK - Looks promising. Will open more options for the testing environment.

Once these changes have been committed, we make changes to the way the exports are listed in test-info.yml for the client machines to consume.

anoopcs9 (approved):
lgtm.

Shwetha-Acharya (approved):

spuiuk (approved):
ACK - the second time (after I just commented and forgot to actually github ack it!)

PR #78, opened 2024-01-18
Provide support for personal and local settings

xhernandez:
With this patch it will be possible to define customized settings using user defined files outside of the repository that will override the values from settings.yml.

This can be useful to adapt sit-environment to the personal environment characteristics without having to modify the settings.yml file each time it's updated.

There are two places were personal settings can be defined:

  - ~/.site/settings.yml
    This file can contain settings used by all sit-environments. This one overrides configurations from settings.yml.

  - ./local.yml (normally the root of the repo)
    This file can contain settings specific for this instance of the repo. This one overrides configurations from settings.yml and ~/.site/settings.yml.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ playbooks/ssh-config-host
 playbooks/Vagrantfile
 playbooks/ansible/config.yml
 devel/authorized_keys
+local.yml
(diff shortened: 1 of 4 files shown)


anoopcs9 (approved):
lgtm, thanks.

PR #82, opened 2024-02-05
smb_share.conf.j2: Add `ignore system acls` option to vfs_acl_xattr

### LINKED ISSUE #81: Possible inclusion of `ignore system acls = yes` option for vfs_acl_xattr
In a recent investigation related to failure with `smb2.session.reauth4` smbtorture test case it was revealed that the `ignore system acls = yes` optional parameter for _vfs_acl_xattr_ module could be a sensible configuration in general to better match NT ACLs provided the lossy translation approach otherwise used. More details to follow..

anoopcs9:
For better coverage of smbtorture tests we require `ignore system acls` enabled for _acl_xattr_ VFS module. This is done to successfully complete those tests where such a lossy translation attempt from NT ACLs to POSIX ACLs would result in failures by design.

fixes #81

### DIFF
--- a/playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
+++ b/playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
@@ -11,4 +11,9 @@ read only = no
 path = {{ subvol }}
 {%- else +%}
 path = {{ path }}
-{%- endif %}
+{%- endif +%}
+{%- if options is defined +%}
+{%- for option, value in options.items() +%}
+{{ option }} = {{ value }}
+{%- endfor +%}
+{%- endif +%}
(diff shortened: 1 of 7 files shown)


xhernandez (commented):
Wouldn't it be better to make it configurable ?

We could also add support for configuring many other options.

playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
```diff
@@ -12,3 +12,8 @@ path = {{ subvol }}
 {%- else +%}
 path = {{ path }}
 {%- endif %}
```
xhernandez, line 14: In this case it doesn't happen, but without the `+` we may lose the new line character to separate options in different lines.

```suggestion
{%- endif +%}
```

xhernandez (approved):

PR #85, opened 2024-02-14
Fixes to adjust with latest ansible v2.16.3

anoopcs9:
Very recently on CentOS Stream 8 ansible version got updated to _v2.16.3_ built with Python 3.12. Due to unavailability of python 3.12 module for _jmespath_ playbook execution is expected to fail. Therefore switch the python module installation from RPMs to pip where it can be versioned.

Additionally due to a regression in `first_found` plugin we switch to absolute paths wherever required to find OS specific files. See https://github.com/ansible/ansible/issues/82695 for more details.

depends on https://github.com/samba-in-kubernetes/samba-centosci/pull/54

### DIFF
--- a/playbooks/ansible/roles/ctdb.setup/tasks/glusterfs/main.yml
+++ b/playbooks/ansible/roles/ctdb.setup/tasks/glusterfs/main.yml
@@ -10,8 +10,10 @@
 
 - name: Process OS specific tasks
   include_tasks: "{{ include_file }}"
+  vars:
+    prefix: "{{ role_path }}/tasks/glusterfs/"
   with_first_found:
-    - files: "{{ config.os[config.nodes[inventory_hostname].os].includes }}"
+    - files: "{{ [prefix] | product(config.os[config.nodes[inventory_hostname].os].includes) | map('join') | list }}"
   loop_control:
     loop_var: include_file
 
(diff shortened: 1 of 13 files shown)


xhernandez (commented):
Instead of adding version dependencies, couldn't we install jmespath using pip instead of the CentOS package ?

anoopcs9:
> Instead of adding version dependencies, couldn't we install jmespath using pip instead of the CentOS package ?

It is my guess that pip executable will also need to used accordingly with required Python version(pip3.11, pip3.12 etc). So its a choice between specifying ansible version and pip executable version. Let me verify that it is the case here.

anoopcs9:
> > Instead of adding version dependencies, couldn't we install jmespath using pip instead of the CentOS package ?
> 
> It is my guess that pip executable will also need to used accordingly with required Python version(pip3.11, pip3.12 etc). So its a choice between specifying ansible version and pip executable version. Let me verify that it is the case here.

Ok. I think I got it confused. I'll check by replacing jmespath python module installation from RPM to pip with required versioned executable.

anoopcs9:
> /retest all

I've manually adjusted required jobs to pick up https://github.com/samba-in-kubernetes/samba-centosci/pull/54 for their execution.

xhernandez (dismissed):
It's a bit ugly but I don't see any other way to do that. Let's make it work for now.

spuiuk:
Isn't playbooks/roles/setup.prep/tasks/centos.yml
being included anywhere? Can we get it to also install the module using pip?

anoopcs9:
> Isn't playbooks/roles/setup.prep/tasks/centos.yml being included anywhere? Can we get it to also install the module using pip?

We discussed offline and decided to have `pip` installation step merged at _playbooks/roles/setup.prep/tasks/centos{8,9}.yml_.

xhernandez (approved):

PR #86, opened 2024-02-21
Normalize share and volume creation

xhernandez:
This patch normalizes the way shares and volumes are created. Till now, the volumes were created during installation of the backend, while shares were created during Samba installation. This caused some difficulties.

Now the volume is created just before creating the share which seems more logical. Some differences between backends have also been removed so that it's more uniform.

This need appeared as part of addressing the comments for #24.


playbooks/ansible/roles/samba.setup/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
 
 - name: Create Samba share directories
   file:
-    path: "{{ config.paths.mount }}/{{ item.cluster_volume }}"
+    path: "{{ config.paths.mount }}/{{ item }}"
```
anoopcs9, line 9: Here `{{ item }}` is a dictionary, right?

xhernandez: Yes

xhernandez: True. I'll fix it.

playbooks/ansible/roles/samba.setup/tasks/main.yml
```diff
@@ -47,28 +47,36 @@
...
-        volume: "{{ item.cluster_volume }}"
-        path: "{{ config.paths.mount }}/{{ item.cluster_volume }}"
-        file: "{{ config.paths.samba.etc }}/smb.shares/{{ name }}.conf"
-        options: "{{ item.samba.options }}"
-      with_items: "{{ samba_shares }}"
+- name: Create mount directories
+  file:
+    path: "{{ config.paths.mount }}/{{ item.name }}"
+    state: directory
+  with_items: "{{ samba_shares }}"
```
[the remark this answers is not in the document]
xhernandez: Yes. I overlooked that one.

anoopcs9 (approved):
lgtm.

Shwetha-Acharya (approved):

PR #90, opened 2024-03-12
Add support for 'gpfs.vfs' backend variant

xhernandez:
Now it's possible to use the 'vfs' variant of the GPFS backend. With it, the gpfs VFS module is automatically configured to be used by Samba.

The GPFS VFS module dynamically loads libgpfs.so library, which requires access to some resources that are not granted by default by SElinux. The Storage Scale product solves this by running smbd process unconfined, so the same solution has been applied here.

### DIFF
--- a/playbooks/ansible/roles/sit.gpfs/templates/smb_share.conf.j2
+++ b/playbooks/ansible/roles/sit.gpfs/templates/smb_share.conf.j2
@@ -1,6 +1,6 @@
 [{{ name }}-{{ config.be.name }}-{{ config.be.variant }}]
 comment = Volume '{{ name }}' from {{ config.be.name }}({{ config.be.variant }})
-vfs objects = acl_xattr
+vfs objects = acl_xattr {%- if config.be.variant == 'vfs' %} gpfs{% endif +%}
 path = {{ path }}
 read only = no
 {%- for option, value in volume.samba.options.items() | default([]) +%}
--- a/playbooks/ansible/roles/samba.setup/tasks/gpfs/main.yml
+++ b/playbooks/ansible/roles/samba.setup/tasks/gpfs/main.yml
@@ -1,1 +1,12 @@
 ---
+- name: Run smbd unconfined
+  when: config.be.variant == 'vfs'
+  block:
+    - name: Configure SElinux context for smbd
+      sefcontext:
+        target: /usr/sbin/smbd
+        setype: bin_t
+        state: present
+
+    - name: Restore SElinux context
+      command: restorecon /usr/sbin/smbd


xhernandez:
This patch doesn't add any specific GPFS configuration, I think most of the default values are already ok, but I'll address that in a future patch. Also, acl_xattr is left enabled for now, but I think GPFS doesn't need it when vfs_gpfs is used.

playbooks/ansible/roles/sit.gpfs/templates/smb_share.conf.j2
```diff
@@ -1,6 +1,6 @@
 [{{ name }}-{{ config.be.name }}-{{ config.be.variant }}]
 comment = Volume '{{ name }}' from {{ config.be.name }}({{ config.be.variant }})
-vfs objects = acl_xattr
+vfs objects = acl_xattr {%- if config.be.variant == 'vfs' %} gpfs{% endif +%}
 path = {{ path }}
```
anoopcs9, line 4: So the `path` value(_/mnt/share_) remains the same and is valid in this case?

xhernandez: Yes. The VFS module in this case is just a helper to provide more features or improve efficiency, but most of the I/O still goes through the kernel mount point.

anoopcs9: > Yes. The VFS module in this case is just a helper to provide more features or improve efficiency, but most of the I/O still goes through the kernel mount point.

Ah.. more or less like _vfs_glusterfs_fuse_ module.

anoopcs9 (approved):

spuiuk (approved):
I initially thought we could use the se boolean samba_run_unconfined instead of relabelling the samba binary. However it appears I was wrong.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/managing_confined_services/sect-managing_confined_services-samba-booleans

samba_run_unconfined
    Having this Boolean enabled allows Samba to run unconfined scripts in the /var/lib/samba/scripts/ directory. 

ACKing these changes.

anoopcs9:
> I initially thought we could use the se boolean samba_run_unconfined instead of relabelling the samba binary. However it appears I was wrong.
> 
> https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/managing_confined_services/sect-managing_confined_services-samba-booleans

On top of it, looking at the following AVC, access is happening via character file.
```
type=AVC msg=audit(1710240306.758:12716): avc:  denied  { read } for pid=102773 comm="smbd[192.168.12" name="ss0" dev="devtmpfs" ino=120415 scontext=system_u:system_r:smbd_t:s0 tcontext=system_u:object_r:gpfs_device_t:s0 tclass=chr_file permissive=0
```

PR #91, opened 2024-03-12
Add 'scale' variant to gpfs backend

xhernandez:


playbooks/ansible/roles/sit.gpfs/tasks/common/main.yml
```diff
@@ -1 +1,5 @@
 ---
+- name: Add GPFS tools to path
+  copy:
+    dest: /etc/profile.d/gpfs-path.sh
+    content: "PATH=${PATH}:/usr/lpp/mmfs/bin"
```
anoopcs9, line 5: Where do we require the above export?

playbooks/ansible/roles/sit.gpfs/tasks/main.yml
```diff
@@ -75,8 +75,54 @@
...
+        - name: Configure shared directory
+          command: >-
+            {{ gpfs_toolkit }} config protocols -f sit_fs -m /gpfs/sit_fs
+              -e {{
+                    config.nodes |
+                    dict2items |
+                    selectattr('value.ctdb.public', 'defined') |
+                    map(attribute='value.ctdb.public') |
+                    join(',')
+              }}
```
anoopcs9, line 97: What are we configuring here?

playbooks/ansible/roles/sit.gpfs/tasks/common/main.yml
xhernandez: This is required later when samba users are created using `smbpasswd` tool, which is installer here by Storage Scale instead of the regular path in `/usr/sbin`

playbooks/ansible/roles/sit.gpfs/tasks/main.yml
xhernandez: This is the set of floating IPs that will be dynamically assigned to the nodes serving SMB protocol. Equivalent to what we normally use for ctdb. That's why I use the same pool of IPs.

playbooks/ansible/roles/sit.gpfs/tasks/common/main.yml
anoopcs9: Ok, got it.

anoopcs9 (approved):

Shwetha-Acharya (approved):

PR #93, opened 2024-03-25
client: install the python3-smbc module

spuiuk:
This python module links to an external C library(libsmbclient) and is therefore very inconvenient to install using pip. We fall back to using the rpm package for this module.

### DIFF
--- a/playbooks/ansible/roles/client.prep/tasks/centos.yml
+++ b/playbooks/ansible/roles/client.prep/tasks/centos.yml
@@ -12,6 +12,7 @@
       - podman
       - tox
       - sshfs
+      - python3-smbc
     state: latest
 
 - name: Link to pytest


anoopcs9:
[pysmbc](https://github.com/hamano/pysmbc) seems to be inactive without any packaging effort in epel repositories.

playbooks/ansible/roles/client.prep/tasks/centos.yml
```diff
@@ -12,6 +12,7 @@
       - podman
       - tox
       - sshfs
+      - python3-smbc
```
anoopcs9, line 15: Which repository provides `python3-smbc`?

spuiuk: ahh sorry. Saw it on my Fedora desktop and assumed it would be available.

The problem here is that we need the python smbc module on the client to allow us to use the smbc python module. The problem is that there is no wheel available which means pip has to compile it. Which requires libsmbclient-devel installed on the system. How do we go about solving this issue without having to install libsmbclient-devel on the client?

spuiuk:
Closing this issue as we are going with pysmb instead of the pysmbc module.

The pysmb python module is a pure python module and doesn't need to be compiled against the libsmb library which means it easier to use in our test environment.

https://github.com/samba-in-kubernetes/sit-test-cases/pull/76

Issue #92, opened 2024-03-25
GPFS sanity tests are failing for `scale` variant


anoopcs9:
All sanity tests seems to fail.

```
=================================== FAILURES ===================================
______________ test_consistency[192.168.123.11-share-gpfs-scale] _______________

ipaddr = '192.168.123.11', share_name = 'share-gpfs-scale'

    @pytest.mark.parametrize(
        "ipaddr,share_name", generate_consistency_check(test_info)
    )
    def test_consistency(ipaddr: str, share_name: str) -> None:
        tmp_root = testhelper.get_tmp_root()
        mount_point = testhelper.get_tmp_mount_point(tmp_root)
>       consistency_check(mount_point, ipaddr, share_name)

testcases/consistency/test_consistency.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

mount_point = PosixPath('/tmp/tmpuli193qc/mnt_jt8yd9j6')
ipaddr = '192.168.123.11', share_name = 'share-gpfs-scale'

    def consistency_check(mount_point: Path, ipaddr: str, share_name: str) -> None:
        mount_params = testhelper.get_mount_parameters(test_info, share_name)
        mount_params["host"] = ipaddr
        try:
            test_file = testhelper.get_tmp_file(mount_point)
            test_file_resp = test_file.with_suffix(".resp")
            test_file_remote = "test-" + ipaddr + "." + share_name
            with open(test_file, "w") as f:
                f.write(test_string)
            put_cmds = "put  %s %s" % (test_file, test_file_remote)
            (ret, output) = testhelper.smbclient(mount_params, put_cmds)
>           assert ret == 0, "Failed to copy file to server"
E           AssertionError: Failed to copy file to server
E           assert 1 == 0

testcases/consistency/test_consistency.py:37: AssertionError
```

See [job page](https://jenkins-samba.apps.ocp.cloud.ci.centos.org/view/GPFS/job/samba_gpfs.scale-integration-environment/) for more details.

PR #95, opened 2024-04-30
Fix share permissions for gpfs.scale

xhernandez:
When the 'scale' variant of gpfs backend is used, the regular samba and ctdb roles are not installed. The permissions to the root directory of the share were defined in these roles.

This patch moves the configuration of the permissions (for all backends) to the new_volume.yml file that is always used to create a new share, even for the 'scale' variant.

Fixes: #92


playbooks/ansible/roles/ctdb.setup/tasks/cephfs/main.yml
```diff
@@ -6,4 +6,3 @@
   vars:
     name: "ctdb"
     path: "{{ ctdb_lock_path }}"
-    mode: "0777"
```
anoopcs9, line 9: So the default _0755_ is enough I guess.

playbooks/ansible/roles/ctdb.setup/tasks/gpfs/main.yml
```diff
@@ -6,4 +6,3 @@
   vars:
     name: "ctdb"
     path: "{{ ctdb_lock_path }}"
-    mode: 0777
```
anoopcs9, line 9: Here also whatever is the default mode should suffice for CTDB cluster lock operations, right?

anoopcs9 (approved):
lgtm, thanks.

playbooks/ansible/roles/ctdb.setup/tasks/cephfs/main.yml
xhernandez: Yes. CTDB runs as root, so there's no need to grant write access permissions to other users.

playbooks/ansible/roles/ctdb.setup/tasks/gpfs/main.yml
xhernandez: Yes. Write access to the owner (root) is enough for CTDB.

PR #96, opened 2024-05-03
Generate newer version of test-info.yml

spuiuk:
The new generation of tests in sit-test-cases require the use of the new version of test-info.yml files. 

Changing the template to generate the new version of test-info.yml files.


xhernandez (commented):
The change looks fine, but I think that the `test-info.yml` creation should be moved inside test.sit-test-cases role, since it's specific to that test. I'm ok if this is done in a future patch, though.

playbooks/ansible/roles/test.sit-test-cases/tasks/prepare/main.yml
```diff
@@ -25,6 +25,15 @@
     version: "{{ test_repo_branch }}"
     dest: /root/sit-test-cases
 
+- name: Create the test-info.yml file with test cluster information
+  template:
+    src: test-info.yml.j2
+    dest: /root/test-info.yml
```
xhernandez, line 31: You could put it directly in `/root/sit-test-cases/test-info.yml` and avoid the symlink creation.

spuiuk: I am trying to remember why I did it like this. I think I did it for re-use. ie. in case other testing projects wanted to make use of the information in that config file.
Another advantage is in a devel environment where we want to delete the whole sit-test-cases repo and redo the folder, it is easier if the configuration is outside the directory.

If it isn't too much of an issue, can we leave it as it is?

spuiuk: Maybe it was a simple case of the repo being pulled after the test-info.yml file was created and hence the conf file wasn't created inside the repo . :) but the points above still hold.

xhernandez: > I am trying to remember why I did it like this. I think I did it for re-use. ie. in case other testing projects wanted to make use of the information in that config file.

The contents of the file are very sit-test-cases specific (that's why I suggested moving it inside the test.sit-test-cases role), so it's unlikely that it will be reused by other tests. Each test can create it's own custom configuration if required.

> Another advantage is in a devel environment where we want to delete the whole sit-test-cases repo and redo the folder, it is easier if the configuration is outside the directory.

You still need to re-run the playbook to recreate the symlink. It's true that you can do it manually, though.

> If it isn't too much of an issue, can we leave it as it is?

Yes, it's not a big problem, and it's contained all inside the `test.sit-test-cases` role, so if you prefer to keep it this way, I'm ok with that.

xhernandez (approved):

PR #97, opened 2024-05-07
sit-xfs: do not recreate fs

spuiuk:
This problem was hit when developing the playbook with backend left to the default xfs system. Attempting to re-run the playbook to test if some changes to the playbooks were working failed at the point of creating the xfs filesystem.

At the moment the playbook forces the creation of the xfs fs. This means that when re-running the ansible playbook, we attempt to recreate a filesystem which has already been mounted. This fails causing it to fail.

By setting to false, we do not force a recreation of the filesystem if one already exists. This makes the operation  idempotent allowing us to re-run the playbook.

Another option is to unmount a mounted filesystem. However it will still wipe out any filesystems on it and is thus non-idempotent.


playbooks/ansible/roles/sit.xfs/tasks/new_volume.yml
```diff
@@ -2,7 +2,7 @@
 - name: Create the XFS filesystem
   filesystem:
     dev: "{{ volume.xfs.device }}"
-    force: true
+    force: false
```
anoopcs9, line 5: `false` is the default in case you don't want to force the creation of filesystem.

spuiuk: I've seen defaults change with ansible. While this setting may not change, I see no harm in explicitly setting it to false.

xhernandez:
> Another option is to unmount a mounted filesystem. However it will still wipe out any filesystems on it and is thus non-idempotent.

I disagree here. If the filesystem is not wiped again, at the end of the playbook execution, the system state won't be the same that the initial execution of the playbook left.

Idempotent means that a given operation, even if executed many times, gives the same result as if it were executed only once. IMO the way to make it idempotent is to actually format it always.

playbooks/ansible/roles/sit.xfs/tasks/new_volume.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: Unmount the filesystem if already mounted
+  mount:
+    path: "{{ path }}"
+    state: unmounted
+
```
xhernandez, line 6: I think it would be better to put this in the caller (playbooks/ansible/roles/samba.setup/tasks/main.yml). This way it will be useful for all backends (though some might need additional steps, but those could be added in future patches)

spuiuk: Xavi, I did some tests with the umount before mkfs. 
I did the following
1) Run "make" which brings up the cluster
2) when the tests are running on the client, I interrupt and
3) rerun make

At this point, I suspect samba still is actively operating on the test filesystem causing the unmount operation to fail. If we wait a while and re-run, it works successfully and completes. 

To get around this, we need to stop samba before we attempt the unmount. Since this is handled by ctdb, it complicates matters.

I think the easier option for now to allow us to re-run make on the cluster is to simply use the option to not force the mkfs for xfs. I have pushed the original patches one more time. Let me know what you think.

xhernandez (approved):

Issue #102, opened 2024-06-03
CentOS Stream 8 reaches EOL


anoopcs9:
A placeholder issue to think about the current usage of CentOS Stream 8 in our environment as it is discontinued with no further updates. As of now GPFS and GlusterFS backends uses CentOS 8 based nodes.

anoopcs9:
@xhernandez Do remember the specific dependency requirement from GPFS side for now EOL CentOS(Stream) 8?

xhernandez:
> @xhernandez Do you remember the specific dependency requirement from GPFS side for now EOL CentOS(Stream) 8?
> 
> Was it some ansible collections?

If I remember correctly, the only issue was that Storage Scale's ansible playbooks were using a deprecated feature (or something that was working differently in the version available in CentOS 9 Stream), causing the scripts to fail installation.

We can try to switch to CentOS 9 Stream to check if the newer installer has fixed the issue. I'll send a patch.

xhernandez:
> We can try to switch to CentOS 9 Stream to check if the newer installer has fixed the issue. I'll send a patch.

It seems there's an issue with the compilation of the GPFS' kernel module.

anoopcs9:
As indicated above GPFS doesn't compile with recent kernel versions from CentOS Stream 9. #104 can be periodically refreshed to check for any developments with upcoming versions.

GlusterFS lacks few _gluster-ansible_ packages on CentOS Stream 9. There is uncertainty in building those missing packages. 

For the time being we are sticking to CentOS Stream 8 with repositories pointing to [vault](https://vault.centos.org). This modification is already in place(#106).

anoopcs9:
> GlusterFS lacks few _gluster-ansible_ packages on CentOS Stream 9. There is uncertainty in building those missing packages.

These are now available via the usual copr repositories. #112 has been created to switch the platform for all GlusterFS related nodes to use CentOS Stream 9.

anoopcs9:
https://github.com/samba-in-kubernetes/sit-environment/pull/104 is now merged and we no longer use CentOS Stream 8 for any backend.

PR #104, opened 2024-06-06
Use default OS for GPFS storage nodes

xhernandez:
This patch removes the requirement of CentOS 8 Stream for the storage nodes of the GPFS backend. That requirement was needed because of an incompatibility of the CentOS 9 Stream's ansible version and the ansible playbooks of the Storage Scale installer.

Updates: #102

### DIFF
--- a/playbooks/settings.yml
+++ b/playbooks/settings.yml
@@ -416,7 +416,7 @@ environments:
       setup:
         cpu_weight: 2
         memory_weight: 2
-        os: centos8
+        kernel: 5.14.0-503
         networks:
           private: 200
           public: 200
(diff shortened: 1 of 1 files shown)


obnoxxx (commented):
the changes LGTM, but wondering if " Updates: #102" in the description and commit message should be "Resolves:" or "Fixes:" ...

xhernandez:
> the changes LGTM, but wondering if " Updates: #102" in the description and commit message should be "Resolves:" or "Fixes:" ...

There are two backends affected by the EOL of CentOS 8 Stream: GPFS and GlusterFS. This PR only tries to fix part of the problem. We'll need to decide what to do with GlusterFS and, maybe, send another patch to fix it. Using "Fixes" would close the issue once the PR is merged, which I don't want. I thought "Updates" is good enough (we used this for multi-patch issues in GlusterFS).

obnoxxx (approved):
LGTM

anoopcs9 (approved):
lgtm.

PR #105, opened 2024-06-06
Expose CephFS snapshots as shadow-copies

Shwetha-Acharya:
Use kernel CephFS mounted share path to create the shadow copy using ceph_snapshots module

Refer: https://www.samba.org/samba/docs/4.11/man-html/vfs_ceph_snapshots.8.html

### DIFF
--- a/playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
+++ b/playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
@@ -1,6 +1,6 @@
 [{{ name }}-{{ config.be.name }}-{{ config.be.variant }}]
 comment = Volume '{{ name }}' from {{ config.be.name }}({{ config.be.variant }})
-vfs objects = acl_xattr
+vfs objects = acl_xattr ceph_snapshots
 {%- if config.be.variant == 'vfs' %} ceph
 ceph:config_file = /etc/ceph/sit.ceph.conf
 ceph:user_id = sit


playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
```diff
@@ -13,3 +13,7 @@ read only = no
 {%- for option, value in volume.samba.options.items() | default([]) +%}
 {{ option }} = {{ value }}
 {%- endfor +%}
+
+[ceph_snapshot]
```
xhernandez, line 17: Shouldn't the snapshots be available through the same share that is used to access the filesystem normally ?

In any case, I think the share name should depend on `name` and `config.be.name` at least to make sure it doesn't conflict with other shares in case the configuration includes more than one share.

playbooks/roles/sit.cephfs/tasks/setup/main.yml
```diff
@@ -7,6 +7,12 @@
       - ansible-collection-community-general
     state: latest
 
+- name: Install samba-vfs-cephfs to support features like vfs_cephs_snapshots
+  yum:
+    name:
+      - samba-vfs-cephfs
+    state: latest
+
```
anoopcs9, line 15: These tasks are included from _setup.prep_ role which is supposed to run only on [admin] nodes. You'll most probably want to install samba-vfs-cephfs on storage nodes.

xhernandez (dismissed):
The change looks fine. The only question is whether we should make the inclusion of `ceph_snapshots` configurable or not, but this can be done in another patch if necessary.

playbooks/ansible/roles/sit.cephfs/tasks/repo/centos.yml
```diff
@@ -3,7 +3,6 @@
   when: inventory_hostname in groups['cluster']
   block:
     - name: Install VFS module
-      when: config.be.variant == 'vfs'
```
anoopcs9, line 6: Can you please add a comment for this task stating the reason for installing `samba-vfs-cephfs` irrespective of the CephFS backend variant?

playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
```diff
@@ -1,6 +1,6 @@
 [{{ name }}-{{ config.be.name }}-{{ config.be.variant }}]
 comment = Volume '{{ name }}' from {{ config.be.name }}({{ config.be.variant }})
-vfs objects = acl_xattr
+vfs objects = acl_xattr ceph_snapshots
```
anoopcs9, line 3: I haven't looked into the internals but can there be a requirement for _ceph_snapshots_ to be placed first in the `vfs objects` list when other modules like _acl_xattr_ etc.(other than _ceph_) are present in the stack?

anoopcs9 (changes requested):

Shwetha-Acharya:
> The change looks fine. The only question is whether we should make the inclusion of `ceph_snapshots` configurable or not, but this can be done in another patch if necessary.

yes that can be done if required.

Shwetha-Acharya:
> The change looks fine. The only question is whether we should make the inclusion of `ceph_snapshots` configurable or not, but this can be done in another patch if necessary.

yes, that can be done if required

playbooks/ansible/roles/sit.cephfs/tasks/repo/centos.yml
Shwetha-Acharya: sure!

Shwetha-Acharya: I was just wondering can there be any possibility of including ceph_snapshots.so as a separate package as othe users with kernel mount essentially would not want to install the vfs module? Or is there any specific reason for this behavior? Please explain if I am missing something. @anoopcs9

anoopcs9: Packaging _ceph_snapshots.so_ as part of `samba-vfs-cephfs` might have stemmed from the fact that it is currently only built conditionally checking the presence of _libcephfs_ headers([source3/wscript](https://git.samba.org/?p=samba.git;a=blob;f=source3/wscript;h=4048ec5aad50c4be3958e53b7c6f4418e8af4301;hb=HEAD#l1914)). But may be we could make adjustments to have it alongside other common vfs modules. I'll modify spec file to reflect such a change. Either we wait for it or install `samba-vfs-cephfs` always for CephFS.

playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
Shwetha-Acharya: I haven't encountered such situation yet. The results with changed order is been same both on the cephfs and cephfs.vfs

playbooks/ansible/roles/sit.cephfs/tasks/repo/centos.yml
Shwetha-Acharya: For now I have updated the comment. If it takes time for the change to come into picture, we can add a new PR to update the changed package

anoopcs9: On a second thought I think its worth waiting for https://github.com/samba-in-kubernetes/samba-build/pull/45 to land so that we avoid the situation with installing `samba-vfs-cephfs` always for CephFS.

Shwetha-Acharya: That also works for me, I will revoke this comment in next change set when the change in samba-build lands

anoopcs9 (changes requested):
I see an odd `z` at the very beginning of the commit subject `zExpose CephFS snapshots as shadow-copies`.

anoopcs9 (approved):
lgtm, thanks.

PR #107, opened 2024-06-13
ceph: Add optional method to use RADOS helper to manage cluster lock

anoopcs9:
It has been a while since a dedicated helper got added to CTDB for managing cluster locks directly on ceph RADOS. Therefore provide a mechanism to optionally configure it instead of the default shared file system locks.

See [man ctdb_mutex_ceph_rados_helper(7)](https://ctdb.samba.org/manpages/ctdb_mutex_ceph_rados_helper.7.html) for more details.

### DIFF
--- a/playbooks/ansible/roles/ctdb.setup/tasks/cephfs/main.yml
+++ b/playbooks/ansible/roles/ctdb.setup/tasks/cephfs/main.yml
@@ -7,3 +7,41 @@
   vars:
     name: "ctdb"
     path: "{{ ctdb_lock_path }}"
+
+- name: Configure RADOS mutex helper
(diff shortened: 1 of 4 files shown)


phlogistonjohn (approved):
lgtm

xhernandez (approved):

anoopcs9:
> This looks very good in general, but I am wondering if we can make rados the default ctdb mutex helper only in the ceph case.

That is already the case. `config.data.ctdb_mutex` is only defined for cephfs backend in [_playbooks/settings.yml_](https://github.com/samba-in-kubernetes/sit-environment/blob/main/playbooks/settings.yml#L358-L364).

PR #109, opened 2024-06-24
samba: Make adjustments to vfs_acl_xattr configuration

anoopcs9:
Following changes are made:
- Remove 'ignore system acls' to keep lossy mapping to POSIX ACLs.
- Configure xattr name to store NT ACLs as "user.NTACL".

We introduce a temporary workaround to allow _smbd_ to assume _cap_dac_override_ when operating as _root_ until official fixes are available.

### DIFF
--- a/playbooks/ansible/cluster-cephfs.yml
+++ b/playbooks/ansible/cluster-cephfs.yml
@@ -22,6 +22,3 @@ samba_netbios_name: "SIT-CEPHFS-TEST"
 
 samba_shares:
   - name: share
-    samba:
-      options:
-        "acl_xattr:ignore system acls": "yes"
--- a/playbooks/ansible/cluster-cephfs.yml
+++ b/playbooks/ansible/cluster-cephfs.yml
@@ -22,3 +22,6 @@ samba_netbios_name: "SIT-CEPHFS-TEST"
 
 samba_shares:
   - name: share
+    samba:
+      options:
+        "acl_xattr:security_acl_name": "user.NTACL"
--- a/playbooks/ansible/cluster-glusterfs.yml
+++ b/playbooks/ansible/cluster-glusterfs.yml
@@ -69,9 +69,6 @@ samba_netbios_name: "SIT-GLUSTERFS-TEST"
 
 samba_shares:
   - name: vol-replicate
-    samba:
-      options:
-        "acl_xattr:ignore system acls": "yes"
     glusterfs:
       type: "replicate"
       subvolume_size: "{{ config.groups['cluster']|length }}"
@@ -90,9 +87,6 @@ samba_shares:
         performance.parallel-readdir: 'on'
 
   - name: vol-disperse
-    samba:
-      options:
-        "acl_xattr:ignore system acls": "yes"
     glusterfs:
       type: "disperse"
       subvolume_size: 3
--- a/playbooks/ansible/cluster-glusterfs.yml
+++ b/playbooks/ansible/cluster-glusterfs.yml
@@ -69,6 +69,9 @@ samba_netbios_name: "SIT-GLUSTERFS-TEST"
 
 samba_shares:
   - name: vol-replicate
+    samba:
+      options:
+        "acl_xattr:security_acl_name": "user.NTACL"
     glusterfs:
       type: "replicate"
       subvolume_size: "{{ config.groups['cluster']|length }}"
@@ -87,6 +90,9 @@ samba_shares:
         performance.parallel-readdir: 'on'
 
   - name: vol-disperse
+    samba:
+      options:
+        "acl_xattr:security_acl_name": "user.NTACL"
     glusterfs:
       type: "disperse"
       subvolume_size: 3
--- a/playbooks/ansible/cluster-gpfs.yml
+++ b/playbooks/ansible/cluster-gpfs.yml
@@ -22,5 +22,8 @@ samba_netbios_name: "SIT-GPFS-TEST"
 
 samba_shares:
   - name: share
+    samba:
+      options:
+        "acl_xattr:security_acl_name": "user.NTACL"
 
 nsd_device: "/dev/vdb"
--- a/playbooks/ansible/cluster-gpfs.yml
+++ b/playbooks/ansible/cluster-gpfs.yml
@@ -22,8 +22,5 @@ samba_netbios_name: "SIT-GPFS-TEST"
 
 samba_shares:
   - name: share
-    samba:
-      options:
-        "acl_xattr:ignore system acls": "yes"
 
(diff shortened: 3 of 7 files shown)


anoopcs9:
With the changes from this PR we have the following failures:

**XFS** and **CephFS kclient**
> test: samba3.**smb2.create.aclfile**
time: 2024-07-04 06:37:01.295294Z
Testing nttrans create with sec_desc on files
basic create
adding a new ACE
creating a file with an initial ACL
time: 2024-07-04 06:37:01.319565Z
failure: samba3.smb2.create.aclfile [
Exception: (../../source4/torture/smb2/create.c:635) Incorrect status NT_STATUS_ACCESS_DENIED - should be NT_STATUS_OK

> test: samba3.**smb2.create.acldir**
time: 2024-07-04 06:37:01.319599Z
Testing nttrans create with sec_desc on directories
basic create
adding a new ACE
creating a file with an initial ACL
time: 2024-07-04 06:37:01.346791Z
failure: samba3.smb2.create.acldir [
Exception: (../../source4/torture/smb2/create.c:635) Incorrect status NT_STATUS_ACCESS_DENIED - should be NT_STATUS_OK

> test: samba3.**smb2.session.reauth4**
time: 2024-07-04 06:32:14.029002Z
time: 2024-07-04 06:32:14.051817Z
failure: samba3.smb2.session.reauth4 [
Exception: ../../source4/torture/smb2/session.c:603: status was NT_STATUS_OBJECT_NAME_NOT_FOUND, expected NT_STATUS_OK: smb2_setinfo_file failed

Earlier with `ignore system acls` set all files and directories created had write permission for others(because samba internally enforces settings like `create mask = 0666`, `directory mask = 0777` etc in this case) and thus _smbd_ didn't have to assume **_cap_dac_override_**(SELinux prevents it though) to perform _fsetxattr(fd, "security.NTACL",...)_ when operating as _root_(**_cap_sys_admin_** was enough(and allowed)).

On the other side when `ignore system acls` is unset(and `acl_xattr:security_acl_name = user.NTACL`) the situation is different where even _root_ will have to override permissions, which is prevented by SELinux at the moment on CentOS Stream 9 packages([fix](https://github.com/fedora-selinux/selinux-policy/pull/1990) to allow _**cap_dac_override**_ is available with later selinux-policy versions), to perform _fsetxattr(fd, "user.NTACL",...)_.

Even with customized local SELinux policy _**smb2.session.reauth4**_(but not others) still remained unsuccessful with the error changed to `NT_STATUS_ACCESS_DENIED`.
> test: samba3.**smb2.session.reauth4**
time: 2024-07-04 06:32:14.029002Z
time: 2024-07-04 06:32:14.051817Z
failure: samba3.smb2.session.reauth4 [
Exception: ../../source4/torture/smb2/session.c:603: status was NT_STATUS_ACCESS_DENIED, expected NT_STATUS_OK: smb2_setinfo_file failed

But this time it failed to set POSIX ACL(**_system.posix_acl_access_**) because the attempt(from _smbd_) was made as Anonymous user(_nobody_) even though we operate using a valid open file handle from the owner. That's one of the places where Samba slightly bend POSIX ACL behaviour in the direction of Windows.

**CephFS vfs**
_libcephfs_ approach survived any of these configuration changes as it seems acting smart enough with open file handles. It includes the situation where _fsetxattr(fd, "security.NTACL",...)_ is allowed for _non-root_ users in a setup with _smbd_ operating without _become_root()_(only **_cap_dac_override_**).

anoopcs9:
I've added an extra commit to temporarily get around the SELinux problems until fixes come in via official packages. So we are left with only the failure from _smb2.session.reauth4_. Since we do not have variant specific flapping list for each backend I'm not sure whether to add _smb2.session.reauth4_ blindly to _flapping.cephfs_ as it will affect both CephFS(kclient) and CephFS(vfs). @spuiuk any thoughts?

xhernandez (approved):

spuiuk:
> But this time it failed to set POSIX ACL(**_system.posix_acl_access_**) because the attempt(from _smbd_) was made as Anonymous user(_nobody_) even though we operate using a valid open file handle from the owner. That's one of the places where Samba slightly bend POSIX ACL behaviour in the direction of Windows.

I thought that this specific problem was only caused on the vfs_ceph2 module which used the anonymouns uid/gid to set UserPerms before making the ceph_ll_setxattr(). vfs_ceph shouldn't really hit this problem with anonymous uid/gid since it totally ignores it and uses the uid/gid setup when initialising the ceph call.

Adding @synarete to this conversation since he seems to be hitting this problem on vfs_ceph_new and didn't have "ignore system acls" set to true in his test setup.

spuiuk:
Let's cover the flapping list issue in the corresponding PR for sit-test-cases
https://github.com/samba-in-kubernetes/sit-test-cases/pull/87

anoopcs9:
> Adding @synarete to this conversation since he seems to be hitting this problem on vfs_ceph_new and didn't have "ignore system acls" set to true in his test setup.

I gave it a try with the following share configuration..
```
[share-cephfs-vfs]
comment = Volume 'share' from cephfs(vfs)
vfs objects = acl_xattr ceph_snapshots ceph_new
ceph_new:config_file = /etc/ceph/sit.ceph.conf
ceph_new:user_id = sit
path = /volumes/_nogroup/share/c5fe416c-5e88-4852-8fb9-8d854f70bfe3
browseable = yes
read only = no
acl_xattr:security_acl_name = user.NTACL
```
..and it turned out to be successful [here](https://jenkins-samba.apps.ocp.cloud.ci.centos.org/view/GitLab/job/samba_gitlab-cephfs.vfs.new-integration/22/).

spuiuk:
I am curious about why it fails with ceph kernel mount. The uid/gid at the time of the operation would be set to nobody/nobody but the open fd would have been opened as the user and the effective uid of the process is set to 0. That should have worked in my opinion.

anoopcs9:
> I am curious about why it fails with ceph kernel mount. The uid/gid at the time of the operation would be set to nobody/nobody but the open fd would have been opened as the user and the effective uid of the process is set to 0. That should have worked in my opinion.

As I explained above https://github.com/samba-in-kubernetes/sit-environment/pull/109#issuecomment-2208737127 the failure happens when we try to set POSIX ACL(**_system.posix_acl_access_**) as Anonymous user(**not** as _root_). In effect _smbd_ performs the operation as _nobody_ and the file system lacks write permission for anyone but owner. Thus we face `EACCES` which is valid from kernel perspective. We are stuck with the exact same situation for XFS and that's why we need a distinction in the flapping list for variants for a backend.

spuiuk (approved):
Just had a call with Anoop where he described the problem in more detail.  Makes sense now.

ACK

PR #111, opened 2024-07-05
Few improvements to test-info.yml

anoopcs9:
- Remove an unnecessary variable definition.
- Add variant in addition to backend.

depends on https://github.com/samba-in-kubernetes/sit-test-cases/pull/89


playbooks/ansible/roles/test.sit-test-cases/templates/test-info.yml.j2
```diff
@@ -10,11 +10,13 @@ users:
     {%- endfor +%}
   {%- endfor +%}
 
-backend: {{ test_backend }}
+backend: {{ config.be.name }}
```
spuiuk, line 13: We should consider removing this global backend valuesince this value will be overridden by the backend-name for each share later.  But this is not a high prio requirement and it may be useful when adding multiple variants in the same test.

spuiuk (approved):
ack

PR #116, opened 2024-08-05
sit.cephfs: Add share configured with `vfs_ceph_new`(non-mgr)

anoopcs9:
depends on #117

### DIFF
--- a/playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
+++ b/playbooks/ansible/roles/sit.cephfs/templates/smb_share.conf.j2
@@ -3,7 +3,10 @@
 comment = Volume '{{ volume.name }}' from {{ config.be.name }}({{ config.be.variant }} {{ method }})
 vfs objects = acl_xattr ceph_snapshots
   {%- if method != 'kclient' %}
-    {%- set vfs = 'ceph' %} {{ vfs }}
+    {%- set vfs = 'ceph_new' %}
+    {%- if method == 'vfs' %}
+      {%- set vfs = 'ceph' %}
+    {%- endif %} {{ vfs }}
 {{ vfs }}:config_file = /etc/ceph/sit.ceph.conf
 {{ vfs }}:user_id = sit
 path = {{ subvol }}
--- a/playbooks/settings.yml
+++ b/playbooks/settings.yml
@@ -362,7 +362,7 @@ environments:
     data:
       branch: main
       ctdb_mutex: rados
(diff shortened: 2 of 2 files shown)


spuiuk:
I think instead of adding a new variable, we should just add to the list of variants.  "vfs_new"?

anoopcs9:
> I think instead of adding a new variable, we should just add to the list of variants. "vfs_new"?

There are multiple aspects to be considered here:

- As of now smb mgr module based configuration is based on `backend=cephfs.mgr.vfs` which indicates the need for ceph native method using the second part of the backend specification. This looked more relatable and meaningful.
- Going forward we may include another configuration using proxy. If go with _vfs_new_ we end up with _vfs_new_proxy_. I think _cephfs.vfs.new_proxy_ will be more sensible. Of course this will end up as _cephfs.mgr.vfs_new_proxy_ for the other case.
- Sooner or later we would have only two: _cephfs_ and _cephfs.mgr_. All other variations are supposed to be clubbed together. Till then it might be easier to put anything extra(new vfs module, proxy requirement) other than the setup method choices separately to avoid any complication in intelligently deriving everything from a single variable.

Let me know if you had something else in your mind.

anoopcs9:
> > I think instead of adding a new variable, we should just add to the list of variants. "vfs_new"?
> 
> There are multiple aspects to be considered here:
> 
>     * As of now smb mgr module based configuration is based on `backend=cephfs.mgr.vfs` which indicates the need for ceph native method using the second part of the backend specification. This looked more relatable and meaningful.
> 
>     * Going forward we may include another configuration using proxy. If go with _vfs_new_ we end up with _vfs_new_proxy_. I think _cephfs.vfs.new_proxy_ will be more sensible. Of course this will end up as _cephfs.mgr.vfs_new_proxy_ for the other case.
> 
>     * Sooner or later we would have only two: _cephfs_ and _cephfs.mgr_. All other variations are supposed to be clubbed together. Till then it might be easier to put anything extra(new vfs module, proxy requirement) other than the setup method choices separately to avoid any complication in intelligently deriving everything from a single variable.

I think I'll reverse the order here to first make the change to group possible shares(starting with cephfs) and then rework this PR to add another share(using _vfs_ceph_new_) on top of it.

spuiuk:
I think the code is far too complex now with a number of if conditions causing it to split into multiple code paths. This change adds another such conditions. I think it would be far easier to go straight to grouping now.

xhernandez (dismissed):

xhernandez (approved):

PR #121, opened 2024-08-30
test-info.yml: Replace `variant` with `method`

anoopcs9:
`variant` is no longer relevant with sit-test-cases as it calculates the flapping list based on the access method used in each of the shares. So specify the `method` for each share section.


Shwetha-Acharya (approved):

xhernandez:
@anoopcs9 the contents of test-info.yml have changed substantially. Is this change still required ?

anoopcs9:
> @anoopcs9 the contents of test-info.yml have changed substantially. Is this change still required ?

I searched through the sheer number of changes made over the past two years, and I think this could still be useful in very specific cases. However, it would require close coordination with corresponding changes in the sit-test-cases repository.

I agree that the meaning of `variant` has changed over time: it now indicates whether the backend can orchestrate the Samba/CTDB setup. Initially, I proposed explicitly specifying the `method` (_vfs_ or _kclient_) in _test-info.yml_, indicating how each share in a variant should be configured. This would allow us to distinguish tests that are expected to pass or fail for specific reasons and list them in separate files in sit-test-cases.

That distinction still applies, but by default, sit-environment no longer exposes _kclient_ based shares to the testing side. Currently, when `variant` is not _default_, sit-test-cases searches for _flapping.cephfs-mgr_ but cannot find it because the actual filename is _flapping.cephfs-vfs_. As a result, tests that must pass on VFS shares are not checked for their outcome. Although they currently complete successfully, any future failures caused by legitimate changes would go unreported.

That said, this could also be addressed without changing `sit-environment`: we could pattern match the share name and derive the corresponding `flapping` filename accordingly.

Issue #123, opened 2024-09-26
GPFS: CI jobs are failing due to lack of minimum required Python version


anoopcs9:
Since the update to v5.2.1.0 our CI jobs started failing with the following error:

```
The IBM Storage Scale requires Python version 3.8 or above, Install the required python version and then retry.
Current version is 3.6.8
```

Please note that we are still stuck with CentOS Stream 8 for GPFS. See https://github.com/samba-in-kubernetes/sit-environment/issues/102.

PR #124, opened 2024-09-27
gpfs: Use python 3.9 in CentOS 8

xhernandez:
The most recent release of GPFS (5.2.1.0) requieres python 3.8+. Since this version doesn't still work with CentOS Stream 9, we install the highest python version available in CentOS 8.

Fixes: #123


anoopcs9 (approved):
Glad that we don't have any ansible issues.

lgtm, thanks.

PR #129, opened 2025-02-10
settings.yml: Use CentOS provided vagrant boxes

anoopcs9:
We previously [switched](https://github.com/samba-in-kubernetes/sit-environment/commit/d1a8534b10b08c0bca28e3ff4445181f8be5fbba) from official CentOS vagrant boxes as they were not regularly updated. But the situation has improved a lot since the last year where boxes are more frequently updated. On top of all that vagrant cloud platform(HCP-Hashicorp Cloud Platform) has recently implemented [rate limits](https://github.com/hashicorp/vagrant/issues/13590) for downloading boxes. Given all these facts it is better to get back to official CentOS vagrant boxes. These are externally hosted on CentOS cloud(and properly redirected from HCP) allowing us to bypass the throttling seen with HCP.

### DIFF
--- a/playbooks/settings.yml
+++ b/playbooks/settings.yml
@@ -189,8 +189,8 @@ tests:
 provisioners:
   vagrant:
     images:
-      centos8: generic/centos8s
-      centos9: generic/centos9s
+      centos8: centos/stream8
+      centos9: centos/stream9
 
     networks:
       private: 192.168.122.0/24


phlogistonjohn (commented):
seems reasonable to me.

anoopcs9:
> We previously [switched](https://github.com/samba-in-kubernetes/sit-environment/commit/d1a8534b10b08c0bca28e3ff4445181f8be5fbba) from official CentOS vagrant boxes as they were not regularly updated. But the situation has improved a lot since the last year where boxes are more frequently updated. On top of all that vagrant cloud platform(HCP-Hashicorp Cloud Platform) has recently implemented [rate limits](https://github.com/hashicorp/vagrant/issues/13590) for downloading boxes. Given all these facts it is better to get back to official CentOS vagrant boxes. These are externally hosted on CentOS cloud(and properly redirected from HCP) allowing us to bypass the throttling seen with HCP.

Hashicorp [confirms](https://github.com/hashicorp/vagrant/issues/13590#issuecomment-2648580244) that it was misconfiguration. Anyway I'll try to fix the CI failures with the switch to CentOS official boxes.

xhernandez:
It looks like the new images already have a newer kernel than the one we want to install and it doesn't get replaced. We should force the installation of the kernel, even if it's a downgrade. I'll try to look at this, maybe tomorrow.

PR #130, opened 2025-02-28
sit.cephfs: Add proxy enbaled shares with `vfs_ceph_new` (mgr)

anoopcs9:
With https://github.com/ceph/ceph/pull/59740 merged ceph smb manager module is capable of creating shares with `ceph_new:proxy` module option set to "yes". Thus an additional share with proxy support is created for _mgr_ variant.


xhernandez (approved):

phlogistonjohn (approved):
I'm a bit rusty wrt how things work in SIT, but seems OK to me.

PR #131, opened 2025-03-17
ctdb.setup: Skip registering rados mutex helper as a service

anoopcs9:
With [!3998](https://gitlab.com/samba-team/samba/-/merge_requests/3998) merged we have an option to not register ctdb rados mutex helper as a service which avoids triggering health warnings as it is not created by cephadm.


xhernandez (approved):

anoopcs9:
CephFS jobs are still failing due to https://github.com/samba-in-kubernetes/sit-test-cases/issues/99. For now ignoring those failures to get us move forward step by step.

PR #137, opened 2025-06-13
hosts.update: Add option to install specific kernel

xhernandez:


playbooks/roles/hosts.update/tasks/centos.yml
```diff
@@ -1,12 +1,62 @@
...
+      loop:
+        - ""
+        - "-core"
+        - "-modules"
+        - "-modules-core"
+        - "-devel"
+        - "-headers"
+
+    - name: Create local repository
+      command: createrepo /root/repo
```
anoopcs9, line 32: I think we can go for a direct install of required (specific) kernel packages from url using dnf/yum without the creation of local repository.

anoopcs9 (commented):
I hope you wanted to see the results early and thus included the second commit rather than reusing https://github.com/samba-in-kubernetes/sit-environment/pull/104 for defaulting to CentOS Stream 9 for GPFS.

xhernandez:
> I hope you wanted to see the results early and thus included the second commit rather than reusing #104 for defaulting to CentOS Stream 9 for GPFS.

Yes, it's mostly a test PR. I ran it locally and it worked, and I wanted to check if it also works in the CI. When the changes are ready, I'll just send the first commit.

playbooks/roles/hosts.update/tasks/centos.yml
anoopcs9, line 32: nit:
```
$ ls -l /usr/bin/createrepo
lrwxrwxrwx. 1 root root 12 Jan 16 05:30 /usr/bin/createrepo -> createrepo_c
```

```suggestion
      command: createrepo_c /root/repo
```

xhernandez: Done

PR #139, opened 2025-07-22
test-info: Include share information from individual files

anoopcs9:
With increasing number of variations to how shares for different backends are populated, the template for _test-info.yml_ has become convoluted. Moreover it lost its generic nature due to additional checks on various backend specific options while forming the shares section. Thus we split the shares section in to separate backend specific templates and include it in the main _test-info.yml_.

### DIFF
--- a/playbooks/ansible/roles/test.sit-test-cases/templates/cephfs-shares-info.yml.j2
+++ b/playbooks/ansible/roles/test.sit-test-cases/templates/cephfs-shares-info.yml.j2
@@ -1,0 +1,14 @@
+shares:
+  {%- for share in samba_shares +%}
+    {%- if test_casesensitive_cephfs is defined or not (share.cephfs.options.casesensitive | default(false)) +%}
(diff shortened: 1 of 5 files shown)


spuiuk:
Should we disable glusterfs testing in the CI?

xhernandez (approved):
I'm wondering if we shouldn't move some of this jinja2 code into configuration options so that templates get cleaner without (much) program logic, but I don't see a clear way to do it right now, so this looks fine to me.

anoopcs9:
> > Should we disable glusterfs testing in the CI?
> 
> We could. Nightly runs are already disabled. For sit-test-cases, it is only triggered with specific label (`centos-ci/glusterfs`). May be we implement similar rule for sit-environment.

This change is in effect now. Hereon GlusterFS tests shouldn't be automatically triggered.

Now [wiki](https://github.com/samba-in-kubernetes/sit-environment/wiki/CentOS-CI) needs some update 😄.

PR #152, opened 2026-04-30
Run all playbooks locally from a virtual python environment

xhernandez:
Instead of creating an additional VM (setup0) to run many of the environment preparation playbooks, which uses valuable resources and is not actively used by many backends, I create a virtual python environment from which all the playbooks are executed locally.

This PR does the following:

- Fix some issues with different versions of ansible
- Fix some deprecation warnings
- Unify all the roles into a single directory (playbooks/ansible is removed)
- Create a python virtual environment for each site
- Move all configuration inside the python's virtual environment
- Cleanup unnecessary setup nodes from the environments
- Unify backend roles inside a single 'backend' role
- Dynamically load the configuration using the 'load.config' role
- Merge all playbooks into a single one and use tags to select the executed plays

This change also has in mind two future extensions:

- The possibility of having multiple test environments simultaneously

  There are more changes required to support this, but this PR already allows the creation of independent sites, each with its own configuration. With multiple sites defined at the same time, it would be easy and quick to switch between them.

- The creation of additional VMs to offer AD support

  By removing setup0, there will be more resources available for these new VM's


xhernandez:
This change passes for all backends, including glusterfs, but I've had to remove the disperse volume since it hangs for some reason. This is the log from samba:

```
[2026/05/08 09:10:25.098182, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/smb2_service.c:71(set_conn_connectpath)
  set_conn_connectpath: service vol-replicate-glusterfs-default, connectpath = /
[2026/05/08 09:10:25.098342, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:511(openat_pathref_fsp)
  openat_pathref_fsp: smb_fname [/]
[2026/05/08 09:10:25.098426, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:415(openat_pathref_fullname)
  openat_pathref_fullname: smb_fname [/]
[2026/05/08 09:10:25.098473,  5, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:77(fsp_new)
  fsp_new: allocated files structure (1 used)
[2026/05/08 09:10:25.098529, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:2530(file_name_hash)
  file_name_hash: / hash 0xa2ff470e
[2026/05/08 09:10:25.098660,  5, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:77(fsp_new)
  fsp_new: allocated files structure (2 used)
[2026/05/08 09:10:25.098707, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:2530(file_name_hash)
  file_name_hash: / hash 0xa2ff470e
[2026/05/08 09:10:25.098778,  5, pid=56472, effective(0, 0), real(0, 0)] ../../source3/modules/vfs_glusterfs.c:640(vfs_gluster_fetch_glfd)
  vfs_gluster_fetch_glfd: Failed to fetch fsp extension
[2026/05/08 09:10:25.098823,  0, pid=56472, effective(0, 0), real(0, 0)] ../../source3/modules/vfs_glusterfs.c:808(vfs_gluster_openat)
  vfs_gluster_openat: Failed to fetch gluster fd
[2026/05/08 09:10:25.098914, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:650(openat_pathref_fsp_simple_openat)
  openat_pathref_fsp_simple_openat: smb_vfs_openat(///) failed: Success
[2026/05/08 09:10:25.098980,  5, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:2420(file_free)
  file_free: freed files structure 0 (1 used)
[2026/05/08 09:10:25.099025, 10, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:487(openat_pathref_fullname)
  openat_pathref_fullname: Opening pathref for [/] failed: NT_STATUS_UNSUCCESSFUL
[2026/05/08 09:10:25.099085,  5, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:2420(file_free)
  file_free: freed files structure 0 (0 used)
[2026/05/08 09:10:25.099127,  3, pid=56472, effective(0, 0), real(0, 0)] ../../source3/smbd/files.c:1851(synthetic_pathref)
  synthetic_pathref: opening [/] failed
[2026/05/08 09:10:25.099171, 10, pid=56472, effective(0, 0), real(0, 0), class=vfs] ../../source3/smbd/vfs.c:1130(vfs_get_fs_capabilities)
  vfs_get_fs_capabilities: Could not open share root
```

Apparently it fails to get the glfs_fd_t associated with the parent fsp of the root of the volume. It would seem an issue on the vfs module, but I don't know why it works on a replicated volume.

Additionally, the gluster's ansible roles have issues with more recent versions of ansible. For now, since the CI is using python 3.9, which doesn't support latest ansible, it's still working. Most probably when we switch to CentOS 10 or a more recent python version it will break gluster's ansible roles.

ansible/Makefile.site
```diff
@@ -0,0 +1,65 @@
...
+ifneq ($(filter-out help,$(MAKECMDGOALS)),)
+ifndef VIRTUAL_ENV
+$(error Site's Makefile must be called inside a Python virtual environment)
+endif
+endif
+
+.PHONY: test
+test: setup.site
+
+.PHONY: statdeump
```
anoopcs9:
```suggestion
.PHONY: statedump
```

Shwetha-Acharya:
lgtm

Shwetha-Acharya (dismissed):

Makefile
```diff
@@ -1,13 +1,158 @@
...
+endef
+export HELP_TEXT
+
+# Enforce a single goal (at most) in the command line
+ifneq ($(word 2,$(MAKECMDGOALS)),)
+$(error Only one goal is allowed)
+endif
+
+# Disable builtin rules and prevent printing of directory changes
+MAKEFLAGS += --no-builtin-rules --no-print-directory
```
anoopcs9, line 37: What calls for `--no-builtin-rules`?

Makefile
```diff
@@ -1,13 +1,158 @@
...
+# path.
+findinpath = \
+	$(strip \
+		$(or \
+			$(firstword \
+				$(wildcard $(addsuffix /$(1),$(subst :, ,$(PATH)))) \
+			), \
+			$(error Command '$(1)' not found in PATH) \
+		) \
+	)
```
anoopcs9: Why can't we make use of shell built-in `command`?

Makefile
```diff
@@ -1,13 +1,162 @@
...
+					$(error Trying to remove '$(notdir $(1))' while in use), \
+					$(notdir $(1)) \
+				) \
+			), \
+			$(error Invalid python venv '$(1)') \
+		) \
+	)
+
+# Determine parent directory of the current Makefile
+MFD := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
```
anoopcs9, line 58: Can we stick to `realpath`?

```suggestion
# Determine parent directory of the current Makefile
MFD := $(patsubst %/,%,$(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
```

anoopcs9 (changes requested):

Makefile
xhernandez: It's to avoid some internal rules that `make` provides mostly for building programs. With them, there's a lot of noise if you want to run `make -d` to see what `make` is actually doing. It also avoids some potential issues in case one of the internal rules matches the special targets.

Makefile
xhernandez: I tried to avoid external dependencies, but I think `command` could be good enough.

Makefile
xhernandez: No. The behavior is not the same. `realpath` resolves the full path by replacing symlinks. Since the Makefile stored in the **site/<site_name>** directory is a symlink, I would always get the directory of the main Makefile (same as `DIR` macro), but I want to get the directory from which the Makefile is executing, which could be the main directory or one of the site subdirectories. `abspath` just combines the path with the current working directory and cleans up it by removing double "/" or resolving "..", "./", but without replacing symlinks.

Makefile
anoopcs9: OK.

Makefile
anoopcs9: OK, understood.

Makefile
```diff
@@ -1,13 +1,144 @@
...
+.PHONY: all clean
+GOALS := all $(filter-out $(GLOBAL_GOALS),$(MAKECMDGOALS))
+
+$(GOALS): site
+ifndef VIRTUAL_ENV
+	@source "$(SITE_DIR)/bin/activate" && \
+		$(MAKE) -C "$(SITE_DIR)" $(MAKECMDGOALS)
+else
+	@cd "$(SITE_DIR)" && $(MAKE) -f "$(BASE)/Makefile.site" $(MAKECMDGOALS)
+endif
```
anoopcs9: I'm still trying to understand this logic. Why do we need this extra level of recursion when activating from scratch? By this point, `SITE` is already determined and the virtual environment has already been created (via the _site_ prerequisite), so we gain nothing from another round of initialization. If reactivating an already activated virtual environment causes no harm, could the following work?

```diff --git a/Makefile b/Makefile
index 292019e..40d0ee0 100644
--- a/Makefile
+++ b/Makefile
@@ -104,12 +104,9 @@ GLOBAL_GOALS := all help site site/$(SITE) $(SITE_DIR) destroy yamllint
 GOALS := all $(filter-out $(GLOBAL_GOALS),$(MAKECMDGOALS))
 
 $(GOALS): site
-ifndef VIRTUAL_ENV
        @source "$(SITE_DIR)/bin/activate" && \
-               $(MAKE) -C "$(SITE_DIR)" $(MAKECMDGOALS)
-else
-       @cd "$(SITE_DIR)" && $(MAKE) -f "$(BASE)/Makefile.site" $(MAKECMDGOALS)
-endif
+               cd "$(SITE_DIR)" && \
+               $(MAKE) -f "$(BASE)/Makefile.site" $(MAKECMDGOALS)
 
 .PHONY: help
 help:
```

ansible/Makefile.site
```diff
@@ -0,0 +1,65 @@
+
+define HELP_TEXT
+  test:          Create, build and test a site
+  statedump:     Dump the state of the site
```
anoopcs9:
```suggestion
  test          Create, build and test a site
  statedump     Dump the state of the site
```

anoopcs9 (commented):
As the current logic stands, nothing prevents the `make` invocation from creating the virtual environment when an unknown goal or target is supplied.
```
$ make foobar
Collecting ansible (from -r requirements.txt (line 1))
  Using cached ansible-13.7.0-py3-none-any.whl.metadata (8.1 kB)
Collecting netaddr (from -r requirements.txt (line 2))
  Using cached netaddr-1.3.0-py3-none-any.whl.metadata (5.0 kB)
Collecting passlib (from -r requirements.txt (line 3))
  Using cached passlib-1.7.4-py2.py3-none-any.whl.metadata (1.7 kB)

. . .

Installing collected packages: passlib, resolvelib, PyYAML, pycparser, packaging, netaddr, MarkupSafe, jinja2, cffi, cryptography, ansible-core, ansible
Successfully installed MarkupSafe-3.0.3 PyYAML-6.0.3 ansible-13.7.0 ansible-core-2.20.6 cffi-2.0.0 cryptography-48.0.0 jinja2-3.1.6 netaddr-1.3.0 packaging-26.2 passlib-1.7.4 pycparser-3.0 resolvelib-1.2.1

[notice] A new release of pip is available: 26.0.1 -> 26.1.1
[notice] To update, run: pip install --upgrade pip
make[1]: *** No rule to make target 'foobar'.  Stop.
make: *** [Makefile:107: foobar] Error 2
```
Can we do something to prevent that, so we can validate first and then proceed with the virtual environment creation?

Makefile
xhernandez: Creating the venv doesn't activate it. The key idea here is that the _ansible/Makefile.site_ must always be executed inside a python environment, which will contain the required ansible and python modules to run everything. If make has not been executed inside a virtual environment (i.e. the user has not activated the environment before running make), we need to switch to that environment, activate it and reprocess the _Makefile_, since the conditions have changed (mainly the _Makefile_ path).

The main things I wanted to cover with this approach are the following:

- If the user has activated a venv when running make, use it, even if we are in other directories (unless the directory corresponds to a different venv, which triggers an error because probably the user wanted to run the command in a site different that the active one)
- If the user has not activated any venv, but it's running make from inside the directory of a venv, activate and use that venv
- Otherwise use the venv specified in the command line or the default one

xhernandez:
> As the current logic stands, nothing prevents the `make` invocation from creating the virtual environment when an unknown goal or target is supplied.
>
> [...]
>
> Can we do something to prevent that, so we can validate first and then proceed with the virtual environment creation?

I wanted to avoid duplicating the list of targets in both files, since _ansible/Makefile.site_ could add and remove targets in the future. I see two options:

1. Scan the _ansible/Makefile.site_ file to get the targets and check them in _Makefile_
   This is not trivial since _make_ doesn't provide an easy way to enumerate them yet. It would rely on parsing the make file itself, which can be done but may look a bit overkill.

2. Move all targets to _Makefile_ and remove _ansible/Makefile.site_
   This makes things cleaner and moves all control to a single file.

I think second option will probably be the best one. What do you think ?

Makefile
xhernandez: Oh, I missed that the line that activates the venv is not removed, sorry. I'll check if that works.

anoopcs9:
> > As the current logic stands, nothing prevents the `make` invocation from creating the virtual environment when an unknown goal or target is supplied.
> > [...]
> > Can we do something to prevent that, so we can validate first and then proceed with the virtual environment creation?
> 
> I wanted to avoid duplicating the list of targets in both files, since _ansible/Makefile.site_ could add and remove targets in the future. I see two options:
> 
>     1. Scan the _ansible/Makefile.site_ file to get the targets and check them in _Makefile_
>        This is not trivial since _make_ doesn't provide an easy way to enumerate them yet. It would rely on parsing the make file itself, which can be done but may look a bit overkill.
> 
>     2. Move all targets to _Makefile_ and remove _ansible/Makefile.site_
>        This makes things cleaner and moves all control to a single file.
> 
> 
> I think second option will probably be the best one. What do you think ?

I'm all for it.

.gitignore
```diff
@@ -1,8 +1,3 @@
 *~
-playbooks/.vagrant
-playbooks/ansible/site_inventory
-playbooks/ansible/ssh_config
-playbooks/Vagrantfile
-playbooks/ansible/config.yml
 devel/authorized_keys
-local.yml
+site
```
anoopcs9:
```suggestion
sites
```

Makefile
```diff
@@ -1,13 +1,194 @@
...
+Targets (only one target allowed per invocation):
+  help           Display this help message
+  clean          Cleanup resources used by the current site
+  site           Create Python virtual environment for default site
+  site/<site>    Create Python virtual environment with specified name
+  destroy        Destroy current Python virtual environment
+  test           Create, build and test a site
+  statedump      Dump the state of the site
+endef
+export HELP_TEXT
```
anoopcs9: Do we still need these `export` statements?

anoopcs9 (changes requested):

.gitignore
xhernandez: Oh, I missed it :(. Thanks for caching it.

Makefile
xhernandez: `HELP_TEXT` is used in a shell command, so it needs to be exported so that the text is correctly printed.

anoopcs9: What about others? `BASE`, `SITE` and `SITE_DIR`.

xhernandez: You're right. They were exported because I needed to pass those values to the other Makefile. Now that they are merged, that's not necessary anymore. Thanks !

xhernandez: I've also found a relatively clean way to avoid the export of `HELP_TEXT`. Let me know if you are ok with that change.

anoopcs9: Looks better.

anoopcs9 (commented):
Do we really need the indirection for `test` and `statedump` targets?

Otherwise lgtm.

xhernandez:
> Do we really need the indirection for `test` and `statedump` targets?

It was to keep the same target names as before, but I'm ok to change them. I'll modify it tomorrow.

anoopcs9 (approved):

PR #153, opened 2026-06-04
Add AD support

xhernandez:


ansible/roles/ad.setup/tasks/main.yml
```diff
@@ -0,0 +1,129 @@
...
+                --server-role dc
+                --function-level {{ config.ad.domains[site.domain].functional_level }}
+                {{ '--use-rfc2307' if config.ad.domains[site.domain].rfc2307 is defined }}
+                --option "dns forwarder = {{ dns_forwarder }}"
+                --option "interfaces = {{ site.networks.public }}"
+
+            - name: Rename administrator
+              when: config.ad.domains[site.domain].admin != "Administrator"
+              command: >-
+                samba-tool user rename Administrator "{{ config.ad.domains[site.domain].admin }}"
```
anoopcs9, line 84: I think the syntax is wrong as per [samba-tool(8)](https://www.samba.org/samba/docs/current/man-html/samba-tool.8.html) and failure went unnoticed because `admin` username was always "Administrator".

```suggestion
                samba-tool user rename Administrator --samaccountname "{{ config.ad.domains[site.domain].admin }}"
```

ansible/roles/users.setup/tasks/accounts_ad.yml
```diff
@@ -0,0 +1,47 @@
+---
+- name: Assign RFC2307 IDs to users
+  command: >-
```
anoopcs9, line 3:
```suggestion
- name: Assign RFC2307 IDs to users
  when: domain.rfc2307 is defined
  command: >-
```

ansible/roles/users.setup/tasks/accounts_ad.yml
```diff
@@ -0,0 +1,47 @@
...
+    samba-tool user addunixattrs "{{ item.key }}" {{ item.value.uid }}
+      --username "{{ domain.admin }}"
+      --password "{{ domain.adminpass }}"
+      --gid-number {{ item.value.primary_group | default(item.value.uid) }}
+  with_items: "{{ domain.rfc2307.users | dict2items }}"
+
+- name: Assign RFC2307 IDs to groups
+  command: >-
```
anoopcs9, line 11:
```suggestion
- name: Assign RFC2307 IDs to groups
  when: domain.rfc2307 is defined
  command: >-
```

ansible/roles/backend/tasks/gpfs/main.yml
```diff
@@ -111,9 +114,36 @@
...
+            mode: 0600
+
+        - name: Configure AD authentication
+          when: config.ad.enabled
+          command: >-
+            /usr/lpp/mmfs/bin/mmuserauth service create
+              --data-access-method file
+              --type ad
+              --netbios-name "{{ site.netbios }}"
+              --user-name Administrator
```
anoopcs9, line 137:
```suggestion
              --user-name "{{ config.ad.domains[site.domain].admin }}"
```

ansible/roles/ad.setup/tasks/main.yml
```diff
@@ -0,0 +1,129 @@
...
+            - name: Wait for AD DC to be ready
+              wait_for:
+                port: 389
+                delay: 1
+                timeout: 60
+
+        - name: Create additional AD DC
+          when: (config.ad.domains[site.domain].dc[0] != inventory_hostname) or (config.ad.domains[site.domain].dns is defined)
+          block:
+            - name: Join DCs to domain
```
anoopcs9, line 101: Don't we need to wait for the primary DC to be available?

ansible/roles/ad.setup/tasks/main.yml
```diff
@@ -0,0 +1,129 @@
...
+          block:
+            - name: Open firewall for samba
+              firewalld:
+                service: "{{ item }}"
+                permanent: yes
+                state: enabled
+              with_items:
+                - samba
+                - kerberos
+                - ldap
```
anoopcs9, line 58: Did we miss `dns`?

ansible/settings.yml
```diff
@@ -133,10 +133,116 @@ accounts:
...
+          - grp
         samba: true
+    groups:
+      grp:
+        gid: 1003001
+        instances: 2
+
+# Active Directory configuration:
+#
+# This section defines the available Active Directory (AD) domainsi and
```
anoopcs9, line 149:
```suggestion
# This section defines the available Active Directory (AD) domains and
```

ansible/roles/ad.setup/tasks/main.yml
```diff
@@ -0,0 +1,129 @@
...
+            - name: Create domain
+              command: >-
+                samba-tool domain provision
+                --realm {{ config.ad.domains[site.domain].realm }}
+                --domain {{ config.ad.domains[site.domain].workgroup }}
+                --adminpass '{{ config.ad.domains[site.domain].adminpass }}'
+                --dns-backend SAMBA_INTERNAL
+                --server-role dc
+                --function-level {{ config.ad.domains[site.domain].functional_level }}
+                {{ '--use-rfc2307' if config.ad.domains[site.domain].rfc2307 is defined }}
```
anoopcs9, line 77: For inline conditional, it is better to use the _else_ part.
```suggestion
                {{ '--use-rfc2307' if config.ad.domains[site.domain].rfc2307 is defined else '' }}
```

anoopcs9 (changes requested):
- Are these `samba-tool` commands idempotent? What about `net ads join`? Is it possible to make them idempotent otherwise?
- CTDB manages _smbd_ via the _50.samba_ event script but _winbind_ is started directly through systemd, bypassing CTDB. This means CTDB won't monitor _winbind_ or restart it on failure. The _49.winbind_ event script should be enabled in `ad.member` (not `ctdb.setup`) because the domain join must complete before _winbind_ can start - and `ctdb.setup` runs before `ad.member`. On cluster nodes, _winbind_ should not be enabled in systemd since CTDB owns the boot lifecycle.

xhernandez:
> * Are these `samba-tool` commands idempotent? What about `net ads join`? Is it possible to make them idempotent otherwise?

They are not idempotent right now. I actually thought about it, and it can be done by adding more tasks to check current state and then conditionally run the command, but I decided to not do that since there are already many other operations that are not idempotent.

Maybe a global cleanup of non-idempotent operations should be done in a future PR.

>     * CTDB manages _smbd_ via the _50.samba_ event script but _winbind_ is started directly through systemd, bypassing CTDB. This means CTDB won't monitor _winbind_ or restart it on failure. The _49.winbind_ event script should be enabled in `ad.member` (not `ctdb.setup`) because the domain join must complete before _winbind_ can start - and `ctdb.setup` runs before `ad.member`. On cluster nodes, _winbind_ should not be enabled in systemd since CTDB owns the boot lifecycle.

Sure. I'll do that.

ansible/roles/ad.setup/tasks/main.yml
xhernandez: Done

ansible/roles/ad.setup/tasks/main.yml
xhernandez: Done

ansible/roles/ad.setup/tasks/main.yml
xhernandez: Done

ansible/roles/ad.setup/tasks/main.yml
xhernandez: This is done in line 92. I basically wait for LDAP service to be online. I think this should be enough to add additional DCs.

ansible/roles/backend/tasks/gpfs/main.yml
xhernandez: Done

ansible/roles/users.setup/tasks/accounts_ad.yml
xhernandez: Done

ansible/roles/users.setup/tasks/accounts_ad.yml
xhernandez: Done

ansible/settings.yml
xhernandez: Done

xhernandez:
I've added `do-not-merge` label since the changes include AD configuration by default. This is just for testing. I think we shouldn't have it enabled by default for normal operation.

ansible/roles/ad.member/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Configure AD members
+  when: site.ad == 'member'
+  block:
+    - name: Configure AD cluster members
+      when: inventory_hostname == groups['cluster'][0]
```
anoopcs9:
```suggestion
      when: inventory_hostname in groups['cluster']
```

ansible/roles/ad.member/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+      block:
+        - name: Join the cluster to the AD domain
+          command: >
```
anoopcs9:
```suggestion
           when: inventory_hostname == groups['cluster'][0]
          command: >
```

ansible/roles/ad.member/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+            -U {{ config.ad.domains[site.domain].admin }}%{{ config.ad.domains[site.domain].adminpass }}
+
+        - name: Enable winbind management by ctdb on cluster nodes
+          command: '{{ config.paths.ctdb.bin }}/ctdb event script enable legacy 49.winbind'
+
+        - name: Restart ctdb
+          service: name=ctdb state=restarted enabled=yes
+
+        - name: Ensure cluster is in HEALTHY state
+          command: ctdb nodestatus all
```
anoopcs9:
```suggestion
          run_once: yes
          command: ctdb nodestatus all
```

ansible/roles/local.defaults/templates/config.yml.j2
```diff
@@ -152,6 +201,9 @@ config:
       system_disk: {{ data.system_disk }}
       {%- endif +%}
       disks: {{ data.disks | default([]) }}
+      domain: "{{ dom.name }}"
+      ad: {{ data.ad | default('standalone') }}
+      netbios: {% if 'cluster' in data.groups %}{{ dom.workgroup | default(dom.name.split('.')[0] | upper) }}-CLUSTER{% else %}{{ name | upper }}{{ idx }}{% endif +%}
```
anoopcs9: This generated NetBIOS name can exceed Samba's 15-character limit when SITE produces a workgroup longer than 7 characters. The default `SITE=default` is valid (DEFAULT-CLUSTER, exactly 15), but longer custom site names cause `net ads join` to fail.

ansible/roles/samba.install/tasks/main.yml
```diff
@@ -6,3 +6,95 @@
...
+    smb_conf: >-
+      {{
+        {
+          "idmap config * : backend": "autorid",
+          "idmap config * : range": config.ad.idmap.min ~ "-" ~ (config.ad.idmap.min + config.ad.idmap.size - 1),
+          "idmap config * : rangesize": config.ad.idmap.block
+        } | combine(config.ad.domains[site.domain].rfc2307 is defined | ternary({
+              "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : backend": "ad",
+              "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : schema_mode": "rfc2307",
+              "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : unix_primary_group": config.ad.domains[site.domain].rfc2307.unix_primary_group | ternary("yes", "no"),
```
anoopcs9: _ternary()_ eagerly evaluates the rfc2307 dictionary, so domains without optional rfc2307 settings fail on undefined _min_, _size_ or _unix_primary_group_.

ansible/roles/local.defaults/templates/config.yml.j2
```diff
@@ -27,62 +27,110 @@ config:
...
+        dc:
+      {%- for name in settings.environments[be].nodes +%}
+        {%- set node = settings.environments[be].nodes[name] +%}
+        {%- if (settings.ad.domains[node.domain | default(domain)].name == dom.name) and (node.ad is defined) and (node.ad == 'dc') +%}
+          {%- for idx in range(node.instances | default(1)) +%}
+          - {{ name }}{{ idx }}
+          {%- endfor +%}
+        {%- endif +%}
+      {%- endfor +%}
+    {%- endfor %}
```
anoopcs9, line 133: When a domain uses a custom admin name, the built-in `Administrator` account is renamed, but the rfc2307 user entry still targets `Administrator`. We have to update the generated rfc2307 mapping to use the configured administrator name for provisioned domains.

anoopcs9 (changes requested):
`cephfs.mgr` still uses `auth_mode: user`, so this variant does not exercise AD authentication. This should use "active-directory" when AD is enabled, with related paths such as AD user creation and test username qualification updated accordingly. The currently passing "cephfs.mgr" tests validate local `auth_mode: user` authentication, not AD membership.

ansible/roles/samba.install/tasks/main.yml
xhernandez: Done

ansible/roles/local.defaults/templates/config.yml.j2
xhernandez: I don't think this needs to be changed. This comes from user-controlled configuration (_settings.yml_). If the user defines a domain whose administrator user has a custom name, then "admin" should be set to indicate that name, and the user specified inside "users" should already use the custom name.

xhernandez:
> `cephfs.mgr` still uses `auth_mode: user`, so this variant does not exercise AD authentication. This should use "active-directory" when AD is enabled, with related paths such as AD user creation and test username qualification updated accordingly. The currently passing "cephfs.mgr" tests validate local `auth_mode: user` authentication, not AD membership.

Actually, I did that on purpose because I didn't have a clear view of the required changes and I preferred to focus on the basic support, though I didn't comment on it. Currently, even if AD support is enabled, machines still behave as 'standalone' servers for the 'mgr' variant.

Anyway, I think I could add it now. I'll try.

ansible/roles/local.defaults/templates/config.yml.j2
xhernandez: Since this is a user-controlled value, I've switched to "SITE-<name>" to save few characters and I've added a length check as a new task, which causes the execution to terminate if the name is too long.

ansible/roles/local.defaults/templates/config.yml.j2
anoopcs9: Too many config fields 😛, thanks for the clarification.

xhernandez:
I've added a change to make it easier to use/not use AD. Now there's a new EXTRA_VARS value that indicates if AD support is used or not: *use_ad*. By default it's 0.

If it's 0, no AD settings are configured, and the "ad" nodes defined as "dc" (domain controllers) are not created. It uses local users.

If it's 1, AD domain controllers are installed and all AD-related configurations applied. It uses domain users.

In the cephfs.mgr backend, AD support works, but rfc2307 settings are ignored for now.

ansible/roles/ad.member/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+            -U {{ config.ad.domains[site.domain].admin }}%{{ config.ad.domains[site.domain].adminpass }}
+
+        - name: Start winbind on client nodes
+          systemd:
+            name: winbind
+            state: started
+            enabled: yes
+
+    - name: Use winbind for identification and authentication
+      command: authselect select --force winbind
```
anoopcs9, line 48:
```suggestion
    - name: Use winbind for identification and authentication
      when:
        - config.ad.managed or inventory_hostname not in groups['cluster']
      command: authselect select --force winbind
```

ansible/roles/load.config/tasks/main.yml
```diff
@@ -11,3 +11,8 @@
   run_once: true
   debug:
     var: config
+
+- name: Check NetBIOS name length
+  when: site.netbios | length > 15
+  fail:
+    msg: "NetBIOS name of {{ invenroy_hostname }} is too long ({{ site.netbios }})"
```
anoopcs9:
```suggestion
    msg: "NetBIOS name of {{ inventory_hostname }} is too long ({{ site.netbios }})"
```

ansible/roles/local.defaults/templates/config.yml.j2
```diff
@@ -27,62 +27,111 @@ config:
...
             {%- endfor +%}
           {%- endif +%}
-          samba: {{ acct.samba | default('false') }}
+          samba: {{ usr.samba | default('false') }}
       {%- endfor +%}
     {%- endfor +%}
   {%- endfor +%}
 
+  ad:
+    enabled: {{ ad | default(settings.environments[be].ad.enabled | default(false)) }}
```
anoopcs9: In _ansible/roles/local.defaults/tasks/main.yml_ we always define `ad`, so the fallback to `settings.environments[be].ad.enabled` is unreachable.

ansible/roles/local.defaults/templates/config.yml.j2
anoopcs9, line 178: This resource calculation still includes DC nodes when AD is disabled, although those nodes are omitted below. Could we apply the same AD-node condition here so resources are distributed only among VMs that will be created?

ansible/roles/users.setup/tasks/accounts_ad.yml
```diff
@@ -0,0 +1,49 @@
...
+    samba-tool user add "{{ item.key }}" "{{ item.value.password }}"
+      --username "{{ domain.admin }}"
+      --password "{{ domain.adminpass }}"
+      {%- if item.value.comment is defined +%}
+      --description "{{ item.value.comment }}"
+      {%- endif +%}
+      {%- if domain.rfc2307 is defined +%}
+      --uid-number {{ item.value.uid }}
+      --gid-number {{ item.value.primary_group | default(item.value.uid) }}
```
anoopcs9: _accounts.users[].primary_group_ is a group name, but `--gid-number` expects a numeric gid.

ansible/roles/test.sit-test-cases/templates/test-info.yml.j2
anoopcs9, line 4: In AD mode, this iterates over every configured account collection, but only the collection selected by `config.ad.domains[site.domain].accounts` is created in the domain. Could we restrict the generated test users to that collection?

anoopcs9 (changes requested):

ansible/roles/ad.member/tasks/main.yml
xhernandez: Done

ansible/roles/load.config/tasks/main.yml
xhernandez: Done

ansible/roles/local.defaults/templates/config.yml.j2
xhernandez: True. Considering it again, I think it doesn't make sense to have an ad.enabled setting per environment. The global "use_ad" should be good enough. This also simplifies things a bit.

ansible/roles/local.defaults/templates/config.yml.j2
xhernandez: Yes, I already fixed it in my local branch. I'll update it soon.

ansible/roles/test.sit-test-cases/templates/test-info.yml.j2
xhernandez: That's how it already worked for local users before this change. The main problem here is that the list of users must be from the machines of the cluster side, not the client, so we can't just select the users from the domain of the client (the server could be running in a different domain).

Since our configuration is quite simple now, I would prefer to leave it as it was before and maybe address this concern in a future PR.

ansible/roles/users.setup/tasks/accounts_ad.yml
xhernandez: Done

ansible/roles/backend/tasks/cephfs/ceph.smb.cluster.yml.j2
```diff
@@ -27,14 +34,32 @@ resources:
...
+    realm: "{{ config.ad.domains[site.domain].realm }}"
+    join_sources:
+      - source_type: resource
+        ref: join-credentials
+  {%- if config.ad.domains[site.domain].rfc2307 is defined +%}
+  custom_smb_global_options:
+    _allow_customization: i-take-responsibility-for-all-samba-configuration-errors
+    "{{ "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : backend" }}": ad
+    "{{ "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : schema_mode" }}": rfc2307
+    "{{ "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : unix_primary_group" }}": {{ config.ad.domains[site.domain].rfc2307.unix_primary_group | ternary("yes", "no") }}
```
anoopcs9:
```suggestion
    "{{ "idmap config " ~ config.ad.domains[site.domain].workgroup ~ " : unix_primary_group" }}": "{{ config.ad.domains[site.domain].rfc2307.unix_primary_group | ternary("yes", "no") }}"
```

ansible/roles/local.defaults/templates/config.yml.j2
anoopcs9: Did you miss to include the required changes here?

xhernandez: Apparently, yes. Sorry for that. It should be fixed now.

xhernandez:
All jobs passed with AD enabled by default. Now I've updated settings.yml to only use one AD in xfs backend (I used 2 for testing), and `use_ad` is 0 by default now.

anoopcs9 (approved):
lgtm, thanks for your patience.

PR #155, opened 2026-07-08
Switch to CentOS Stream 10 and fix ceph shaman build

anoopcs9:
- Ceph builds on shaman moved from centos/9 to rocky/10, breaking CI. Add a configurable `shaman_distro` field to the cephfs environment data and use it in the shaman URL instead of deriving it from the node OS.
- Add `centos10` _os_info_ and vagrant image mapping, and switch the glusterfs, xfs and cephfs environments to CentOS Stream 10. GPFS remains on CentOS Stream 9.


ansible/settings.yml
```diff
@@ -289,6 +294,7 @@ provisioners:
 #
 environments:
   glusterfs:
+    os: centos10
```
xhernandez, line 297: Why not make centos10 the default ?

xhernandez:
I've executed a query to shaman for CentOS 9 and I got this:

```bash
$ curl -s "https://shaman.ceph.com/api/search/?project=ceph&ref=main&flavor=default&status=ready&distros=centos/9/x86_64&sha1=latest" | jq .
```
```json
[
  {
    "url": "https://1.chacra.ceph.com/r/ceph/main/bf449d4db29d3a1d39d50e20defefc4d6034bc5e/centos/9/flavors/default/",
    "chacra_url": "https://1.chacra.ceph.com/repos/ceph/main/bf449d4db29d3a1d39d50e20defefc4d6034bc5e/centos/9/flavors/default/",
    "ref": "main",
    "sha1": "bf449d4db29d3a1d39d50e20defefc4d6034bc5e",
    "distro": "centos",
    "distro_version": "9",
    "distro_codename": null,
    "modified": "2026-07-14 21:28:56.782967",
    "status": "ready",
    "flavor": "default",
    "project": "ceph",
    "archs": [
      "arm64",
      "source",
      "x86_64"
    ],
    "extra": {
      "version": "21.3.0-1023-gbf449d4d",
      "package_manager_version": "21.3.0-1023.gbf449d4d",
      "build_url": "https://jenkins.ceph.com/job/ceph-dev-pipeline/7200/",
      "root_build_cause": "",
      "node_name": "10.20.192.34+bath04",
      "job_name": "ceph-dev-pipeline"
    }
  }
]
```
It looks like CentOS 9 builds are still provided.

Using a different distro could cause package dependency issues or incompatibilities. If we really need to switch to rocky/10, I think it would be better to change the base OS of the VMs to match the Ceph builds.

ansible/settings.yml
anoopcs9: That's doable, but what about GPFS?

anoopcs9:
> It looks like CentOS 9 builds are still provided.
>
> Using a different distro could cause package dependency issues or incompatibilities. If we really need to switch to rocky/10, I think it would be better to change the base OS of the VMs to match the Ceph builds.

Yeah, this was a byproduct of my initial investigation, which turned out to be wrong later. Otherwise, I had planned to start using CentOS Stream 10 because the Ceph container images (`quay.ceph.io/ceph-ci/ceph:main`) have already switched to Rocky 10, but our nodes are still on CentOS Stream 9. Ideally - and in practice, this mismatch shouldn't cause any problems - but I thought it was worth trying.

Given that, I'm not sure which direction is best: use CentOS Stream 10 or Rocky Linux 10? I'm leaning toward CentOS Stream 10 since it should be largely compatible. For Ceph, everything is containerized, so the underlying distro shouldn't matter, but matching major versions is still preferable.

xhernandez:
IMO, if we don't see issues with CentOS Stream 10, I would prefer to use it.

ansible/settings.yml
xhernandez: In the official documentation it says RHEL 10.x is supported (https://www.ibm.com/docs/en/storage-scale/6.0.1?topic=storage-scale-faq#fsi), so if we install the same kernel version I think it should work.

Right now the latest supported kernel is 6.12.0-211.33.1.el10_2, so using `kernel: 6.12.0-211` in the gpfs environment should avoid the same problems we had with CentOS Stream 9.

anoopcs9: Done

ansible/roles/backend/tasks/cephfs/common/centos.yml
```diff
@@ -1,12 +1,22 @@
 ---
+# shaman publishes ceph builds under a different distro depending on the EL
+# version: EL9 lives under centos/9, but from EL10 on there is no centos
+# build and the packages live under rocky/10. Derive the right distro from
+# the OS set for the node instead of hard-coding it.
+- name: Determine shaman distro for the node OS
```
xhernandez, line 6: This is fine for now, but it could become a mess if we need to create more special cases. If more dependencies like this appear in the future we'll need to consider moving this to some mapping definition in a configuration file.

xhernandez (approved):
The change to dynamic addresses for all interfaces is unfortunate. Right now I don't have a better alternative, so this change is fine, but we lose the ability to quickly identify the nodes by IP that we were using during manual testing/debugging.

In the future, if we can't really use static addresses, I would move all configuration to make it dynamic in all cases.

xhernandez:
Looks like the nightly repos for GlusterFS only contain packages up to CentOS 9. For now I think we should force the `glusterfs` environment to use CentOS 9.

ansible/roles/backend/tasks/cephfs/common/centos.yml
anoopcs9: Agreed. If more OS-specific mappings like this come up, let's move them into a mapping in the config instead of conditionals.

anoopcs9 (commented):
> The change to dynamic addresses for all interfaces is unfortunate. Right now I don't have a better alternative, so this change is fine, but we lose the ability to quickly identify the nodes by IP that we were using during manual testing/debugging.

Agreed, the loss of predictable IPs is unfortunate.

> In the future, if we can't really use static addresses, I would move all configuration to make it dynamic in all cases.

What exactly would you want to make dynamic? Do you mean the CTDB virtual IPs and the per-node host numbers we still set in the config?

anoopcs9:
> Looks like the nightly repos for GlusterFS only contain packages up to CentOS 9. For now I think we should force the `glusterfs` environment to use CentOS 9.

Ah, I was under the impression that sanity checks were always failing with GlusterFS, but that doesn't seem to be the case. Also, please note that we're seeing some CTDB health-check issues even with CentOS Stream 9 nodes - something I may have missed over the past few weeks. I'm looking into it.

xhernandez:
> Ah, I was under the impression that sanity checks were always failing with GlusterFS, but that doesn't seem to be the case. Also, please note that we're seeing some CTDB health-check issues even with CentOS Stream 9 nodes - something I may have missed over the past few weeks. I'm looking into it.

ctdbd says this:

> 2026-09-02T15:36:06.371967+00:00 storage0 ctdbd[17813]: Unknown interface eth2 for public address 192.168.123.10

And I see this in /var/log/messages:

> Sep  2 15:33:58 ip-192-168-121-40 kernel: virtio_net virtio5 ens8: renamed from eth2

So it seems that the CentOS Stream 10 image renames network interfaces.

xhernandez:
> > In the future, if we can't really use static addresses, I would move all configuration to make it dynamic in all cases.
> 
> What exactly would you want to make dynamic? Do you mean the CTDB virtual IPs and the per-node host numbers we still set in the config?

I mean the specific IP configurations defined in each environment could be simplified to a list of networks.

Now we have:

```yaml
   networks:
     private: 100
     public: 100
```

We could just have:

```yaml
   networks:
     - private
     - public
```

But that's something to decide in the future, not for this PR.

anoopcs9:
> > Ah, I was under the impression that sanity checks were always failing with GlusterFS, but that doesn't seem to be the case. Also, please note that we're seeing some CTDB health-check issues even with CentOS Stream 9 nodes - something I may have missed over the past few weeks. I'm looking into it.
> 
> ctdbd says this:
> 
> > 2026-09-02T15:36:06.371967+00:00 storage0 ctdbd[17813]: Unknown interface eth2 for public address 192.168.123.10
> 
> And I see this in /var/log/messages:
> 
> > Sep  2 15:33:58 ip-192-168-121-40 kernel: virtio_net virtio5 ens8: renamed from eth2
> 
> So it seems that the CentOS Stream 10 image renames network interfaces.

This is definitely a problem, but GlusterFS previously had an issue starting the daemon because I had failed to load the required policy introduced with https://github.com/samba-in-kubernetes/sit-environment/pull/157 for GlusterFS. I added an extra commit to address this, and hopefully all backends are now on par.

Still, I'm fairly certain that the interface renaming did not affect the initial runs, since they were successful for all backends except GlusterFS. I'll come up with a solution.

xhernandez (approved):

PR #156, opened 2026-07-10
backend/cephfs: Workaround SSL failures for shaman API

anoopcs9:
The _ansible.builtin.uri_ and _ansible.builtin.get_url_ modules fail against shaman.ceph.com with an SSL error (ASN1: NOT_ENOUGH_DATA, status -1), likely due to the endpoint requiring a higher TLS version than the modules negotiate. Replace _uri_ and _get_url_ with shell/curl tasks as a workaround.

### DIFF
--- a/ansible/roles/backend/tasks/cephfs/common/centos.yml
+++ b/ansible/roles/backend/tasks/cephfs/common/centos.yml
@@ -1,21 +1,16 @@
 ---
 - name: Identify latest ceph:{{ config.data.branch }} build
-  uri:
-    url: "https://shaman.ceph.com/api/search?\
-            project=ceph&\
-            ref={{ config.data.branch }}&\
-            flavor=default&\
-            status=ready&\
-            distros=centos/{{ config.os[site.os].version }}/x86_64&\
-            sha1=latest"
-    return_content: true
-  register: ceph_shaman
+  shell: curl -s "https://shaman.ceph.com/api/search/?project=ceph&ref={{ config.data.branch }}&flavor=default&status=ready&distros=centos/{{ config.os[site.os].version }}/x86_64&sha1=latest"
+  register: get_ceph_shaman
+
+- set_fact:
+    ceph_shaman: "{{ get_ceph_shaman.stdout | from_json }}"
 
 - name: Create ceph repo
   yum_repository:
-    name: "ceph-{{ config.data.branch }}-{{ ceph_shaman.json[0].sha1 }}"
-    baseurl: "{{ ceph_shaman.json[0].url }}/x86_64"
-    description: "Ceph {{ config.data.branch }} ({{ ceph_shaman.json[0].sha1 }})"
+    name: "ceph-{{ config.data.branch }}-{{ ceph_shaman[0].sha1 }}"
+    baseurl: "{{ ceph_shaman[0].url }}/x86_64"
+    description: "Ceph {{ config.data.branch }} ({{ ceph_shaman[0].sha1 }})"
     enabled: true
     gpgcheck: false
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Just my 2c, but shelling out to curl when ansible should be able to make https requests internally seems a bit icky.  Are you stuck on an old version of ansible/python or something?

anoopcs9:
> Just my 2c, but shelling out to curl when ansible should be able to make https requests internally seems a bit icky. Are you stuck on an old version of ansible/python or something?

The pull request https://github.com/samba-in-kubernetes/sit-environment/pull/155 was my initial attempt, based on the (incorrect) belief that we no longer receive _el9_ builds for the Ceph `main` branch. Shortly after, I realized that wasn't the case; the error was actually pointing to an SSL/TLS handshake problem. I then investigated the issue assuming that the minimum TLS requirement had been raised to at least 1.2. Do you know whether there have been any changes to the TLS version requirements at the shaman endpoint?

```
# ansible --version | grep "python version"
  python version = 3.9.25 (main, Apr 17 2026, 00:00:00) [GCC 11.5.0 20240719 (Red Hat 11.5.0-14)] (/usr/bin/python3)

# python3 --version
Python 3.9.25

# openssl --version
OpenSSL 3.5.7 9 Jun 2026 (Library: OpenSSL 3.5.7 9 Jun 2026)
```

I tried a few things to verify that ansible/python can perform a TLS 1.2 (or newer) handshake, but everything seemed fine, so I eventually got fed up and switched to _curl_. Do you have any pointers for what could be failing in the ansible-python integration?

xhernandez:
I agree with John. We already have too many `shell` and `command` tasks, which break many of the benefits of ansible. I would prefer to keep specific tasks if possible.

Python 3.9 is pretty old, and it doesn't support latest ansible version. If I remember correctly, python 3.9 only supports ansible 7 (ansible core 2.14), but latest ansible version is 14 (ansible core 2.21). I think this could be the reason for the problems with TLS 1.2 support.

I think 3.14 is available in CentOS 9, so maybe we can install it in the host machines and run `make PYTHON=python3.14 test`. This should automatically install the latest ansible version, which I guess it should support TLS 1.2.

xhernandez:
I've done some tests with python 3.9 and 3.14, on Fedora and CentOS 9 Stream (the same VM created by sit-environment) and I'm unable to see the failures.

What I've observed is that the shaman certificate was changed very recently (it's valid only from July 7th), and the error captured by the CI failed jobs is:

```
Connection failure: [ASN1: NOT_ENOUGH_DATA] not enough data (_ssl.c:4192)
```

Which looks like a corrupted encoding (Certificates are usually encoded using ASN1). However this should have affected all clients accessing the site, but `curl` was working fine, so the only remaining possibility that I can think of is that ansible or some library used by it had an issue. In this regard, I've found this:

https://github.com/python/cpython/issues/151504
https://github.com/openssl/openssl/pull/31818

Which could be related.

Latest CI runs executed today are passing without any change, so whatever was causing the issue seems to have been fixed already.

I've also tried to force TLS 1.1 to connect to shaman and it fails, so TLS 1.2 must be used to connect there. If the jobs are succeeding now, it means that ansible is also using TLS 1.2 correctly.

anoopcs9:
@xhernandez Thanks for finding the relevant links.

I took this little further and confirmed that we have an updated OpenSSL from today (openssl-3.5.7-**1** to openssl-3.5.7-**2**) in which the only patch included was the fix from https://github.com/openssl/openssl/pull/31818 😊.

This completes the root cause analysis. I'm closing this PR.

PR #158, opened 2026-07-20
samba.setup: Restore smbd_dac_override custom SELinux policy

anoopcs9:
The `smbd_dac_override` policy was removed in commit b839614 assuming it was included in official packages across all supported distributions. However, only CentOS Stream 10 ships this policy by default. Restore it so that other distributions continue to get _cap_dac_override_ allowed for _smbd_t_.

### DIFF
--- a/ansible/roles/samba.setup/files/smbd_dac_override.te
+++ b/ansible/roles/samba.setup/files/smbd_dac_override.te
@@ -1,0 +1,10 @@
+
+module smbd_dac_override 1.0;
+
+require {
+	type smbd_t;
+	class capability dac_override;
+}
+
+#============= smbd_t ==============
+allow smbd_t self:capability dac_override;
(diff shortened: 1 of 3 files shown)


xhernandez (approved):
Are there plans to include it in CentOS Stream 9 standard packages at some point, or that won't happen ?

anoopcs9:
> Are there plans to include it in CentOS Stream 9 standard packages at some point, or that won't happen ?

I think it is hard to make a compelling case for a backport, given the narrow and strict release policy for a Stream version. Currently, Stream 9 is based on v38.1 and Stream 10 on v42.1. The required commit [8d7c223b](https://github.com/fedora‑selinux/selinux‑policy/commit/8d7c223b38ae0bc48b0f8ccc7aaf659b7deee6be) is only present from v40.14 onward. In short, we could try, but it is unlikely to happen for Stream 9.

xhernandez:
Ok. That's fine. I just wanted to know if this would be temporary or it will remain as long as we support CentOS Stream 9. I think it doesn't make sense to try to include it in CentOS Stream 9 just for this reason.

